PR #12, opened 2015-09-04
Datepicker

javivelasco:
Finally, `DatePicker` and `TimePicker` components! A couple of improvements remaining (merging now because I'm sick of it and want to do something else for now):
- Add transition effects in clock and calendar when the view is switched.
- Add limitations properties for max/min values in both pickers.
- Add property to customize allowance of year list picker in date picker.
- Add property to customize auto-transition on select for date picker and time picker.
- Extract to a new module common behaviors, check here slider component too.


javivelasco:
:tada: :tada: :tada: :tada: :tada: :tada:

PR #13, opened 2015-09-22
Rewrite react toolbox in ES6

javivelasco:
More improvements and refactoring coming, but we get rid of coffeescript and write the whole package using ES6 and babel.


soyjavi:
amazing +10000

Issue #16, opened 2015-10-20
How about use es6 classes, and do not use mixin?


ustccjw:
As https://medium.com/@dan_abramov/mixins-are-dead-long-live-higher-order-components-94a0d2f9e750 said, we can use composition instead of mixin.

`react-addons-pure-render-mixin` is not necessary，i do not think we should add it to each component. May be we can use immutable.js.

javivelasco:
First of all thanks for your contribution, you're the first one!

I agree about removing `createClass` and use classes instead but I also would like to keep components as functional as possible. At first we wanted to avoid classes and then I thought about moving to [factories](https://medium.com/javascript-scene/baby-s-first-reaction-2103348eccdd) to keep everything stateless and to use stores to hold data. Also DI is a plus... but we should be pragmatic.

In a component library like this, avoiding state may be not the best choice since you are forcing the consumers to implement stores for every data they want to keep in the application and in their components. This collides with the principle of keeping it simple and making as easy as possible to build something with react toolbox. So yes, we are moving to classes for components that keep state but trying to embrace stateless components from React `0.14` (I know you did in your PR!). What I don't want is to use ES7 notation, too experimental for now. Removing the `PureRenderMixin` is ok too, we were using it to be sure we were not messing up state.

If you remove the ES7 notation I'll be happy to accept the PR, great work!!

ustccjw:
In my option, react-toolbox is forward  and experimental(such as  using `css-modules`). So `stage 0` is always ok.

This PR also use `es7.classProperties(stage 0)`, it is also  in React doc(https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html).

javivelasco:
Property initializers are ok and I agree with you, it's highly experimental and maybe it fits. I don't like the double colon to bind the context though, and in many cases React is autobinding so it can be removed. The property initializers look quite elegant... Do you think it's possible to make eslint work with all this new features? It it's possible and @soyjavi agrees we can merge.

Also we need to solve conflicts before merging since we are working hard to release a stable version ASAP!

ustccjw:
`babel-eslint` makes eslint work with all this new features. 
es6 classes bind is a know issue(https://github.com/facebook/react/issues/3509).

ustccjw:
ok, i think we should use：

> ES7 will likely include class property initializers and you'll then be able to specify your method with an arrow function to autobind it:

class Foo extends React.Component {
  render() {
    // ...
[...]

PR #19, opened 2015-10-21
Move to ES6 classes with ES7 class properties

javivelasco:
This PR resolves what we were talking about in #16. We already have everything migrated to classes but keeping the project with Babel at `stage 2` and including `es7.classProperties` which is now at stage 0. Check [here](https://github.com/jeffmo/es-class-static-properties-and-fields).

The reason is that moving to clases has IMHO two big disadvantages: 
- Ugly declaration of `PropTypes`, `defaultProps` and `initialization`.
- No autobinding for callback functions in JSX.

The first was resolved including class properties which is pretty clever and even [documented](https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html) at React docs as @ustccjw said. The latter was initially resolved with an `autobind decorator` included as a dependency package. BUT we have with ES6 fat arrow functions and **class properties** initialization because of the first mentioned point. We can now take advantage of both and declare event handlers initialized using fat arrow functions like this:

```js
handleMouseDown = (event) => {
  events.pauseEvent(event);
  this.refs.ripple.start(event);
}
```

This is more declarative since you can see the actual context and we can have explicit control over context for each case. It also introduces an organizational pattern leaving in the components the following order: proptypes - defaultprops - state - lifecycle - event handlers - additional functions - rendering functions. So for now: stage 2 + class properties can be a great choice! I think the code now looks way better.

Please @soyjavi, merge this PR once checked and if you're :+1:


ustccjw:
Nice job! I try to use class properties to solve `this bind`, I also notice a problem: 

    state = {
        date: this.props.initialDate,
        display: 'months',
        actions: [
          { label: 'Cancel', className: style.button, onClick: this.onDateCancel.bind(this) },
          { label: 'Ok', className: style.button, onClick: this.onDateSelected.bind(this) }
        ]
    };

if `onDateCancel` use class properties, then onDateCancel declare should be before state declare. But you notice it !

javivelasco:
Yes! I think it's nice this kind of "lazy bind" to solve it, anyway it would be in just a few cases. Awesome! I'm glad you like it :smile:!

soyjavi:
+1

PR #20, opened 2015-10-23
update webpack dev, test and build & update eslint rule

ustccjw:
A lot of changes.(example dir is not modified - may be replaced by site)

1. update dev, test, build config

2. update eslint rule and fix all files

`npm start`: development  mode, open `http://0.0.0.0:8080/spec/`(use react-hot-loader)

`npm run build`: build components to lib

`npm test`: test components

`npm run lint`: use eslint to check js/jsx (maybe add css check)


javivelasco:
Great work! I love it in general terms. Same as two days ago: I've created a branch to test it and saw a couple of problems that I fixed. I'm merging that branch. Just a few comments:

- I want to keep `eslintignore` with `node_modules`, otherwise it takes a lot of time to run and also we are linting code that is not ours. Also I've added some `eslint` rules for ES6 that you added here but keeping the old ones for codestyle consistency. Some of them: always semicolon, no-unused, etc. I know unused can be annoying but it should happen in very limited cases.
- Great ES6 code improvements but you should be careful. There was a bug in the ripple component because of a destructuring change. All changes make a lot of sense.
- Since I'm using regularly node `0.10.x` in my regular job, I've added `.nvmrc` file with the node version which I'm using to run the whole project.
- I added the skipping rules for those cases where eslint complains although it's not bad code. It's better to have the rule with some exceptions that no rule at all IMO.
- I don't think we need `strict` primitives in webpack configuration files, don't we? I can put them back.
- I've upgraded some dependencies.

Great work, thanks :+1:

PR #22, opened 2015-10-24
small fix & mv normalize.scss to local

ustccjw:
just avoid unnecessary problems.


spec/style.scss
```diff
@@ -1,4 +1,12 @@
+@import "../components/commons";
+
 .app {
```
javivelasco: I think this class belongs to the `commons` file. You have to think that every client app using React Toolbox would need to assign this styles to the app root node in order to have properly working Drawer and Dialog components. Why moving it here?

javivelasco:
I agree most of the changes but I would also like to know why moving normalize to the project. I think that's actually a third party dependency that doesn't belong to the project. Also, note that `.app` might belong to `commons.scss` file since it needs to be imported by client apps.

The rest of the changes are ok. Tell me about it and we can merge!

ustccjw:
node-sass will convert `@import "~ normalize.css" `  to `@import url(~ normalize.css)`(lib).

spec/style.scss
ustccjw: Ok, i will make it  recovered.

javivelasco: Maybe in the future this is going to be changed to an `App` component wrapper that already would have the proper styles. This way instead of importing `styles.scss` and add `app` as a className, the user can just make the app hang from `<App />`. We will change it as soon as we write some layout components.

javivelasco:
I don't get why is that happening. The purpose of the import is to dump `normalize.css` in the resulting `CSS` files. It's working for the tests I'd been doing and it should work for you too.😕

ustccjw:
@javivelasco It's sass-loader feature.

Now we should use css-loader in app(because we use css-modules). If we do not convert sass file to css in lib, we will add sass-loader in app.

So  i think css-loader is ok in react component, but sass-loader is not necessary.

javivelasco:
Ok, I get it. Maybe we can better use webpack to generate the `lib` or to do something about. Anyway I'm not sure if we should go with this `lib` strategy. I'd been thinking during this weekend about customization. In an ideal scenario the user should be able to provide a SASS configuration file to override our sass variables. That would be the perfect case to have custom colors.

The only way to get that working is to provide a `react-toolbox-loader` that accepts some configuration from webpack config file to prepend the user configuration to each SASS module file. If we are providing a `lib` based in css modules, why not providing it base in sass modules instead? For the user is just adding other loader to the pipe and it will provide a lot of return.

In that scenario it doesn't make sense to provide a `lib`. I'm planning to write a `react-toolbox` loader soon. What do you think?

ustccjw:
I support `react-toolbox-loader`! It is a good way to custom component style. You can create a repo first!

javivelasco:
Great to know!! :) This afternoon I'm going to create the repo with a first approach. Let's go that way and keep normalize as a module. I think we can remove the `lib` build process to prevent confussion and make `components` the entry point of the package.

The rest of the changes are perfect :+1:

ustccjw:
I think lib is also necessary. We should use babel to convert it to es5, because many app ignore node_modules(babel-loader, for speed).

javivelasco:
I would love to have both options but still sounds weird not having normalize as a dependency. Maybe we can figure out a way to generate `lib` including normalize without having to move it to the repository. It should be a way, always! :dancer:

PR #23, opened 2015-10-29
use react-transform instead of react-hot & fix css-modules composes & add react rules to eslint

ustccjw:
notice: 

1. process.env.UV_THREADPOOL_SIZE should set larger! 

2. react-hot will be Deprecated. (react-transform solution do not support pure function component hot loading, but in progress ).

3. css-modules composes should use sass-loader first.

4. add some react rules to eslint.


javivelasco:
It looks great! Let me test it and I'll merge

javivelasco:
Done, great!
Do you have feedback for the documentation?

ustccjw:
I like the structure. I will watch it carefully!

PR #26, opened 2015-10-30
add sass sourcemap in dev & add app_bar to spec & fix lint

ustccjw:


ustccjw:
@javivelasco how do you think stateless component? I want to hear your pointer.

For example, `Input component` has a `value` state, I think dump component's state is some pointer that we don't care - we can call it `temporary`.

javivelasco:
It's a topic I'm discussing a lot with @soyjavi. I'd like to get as much feedback as possible. Since I'm not yet using React Toolbox components in a large scenario I don't know how valuable is to have `state` in some components that actually can be stateless... The general rule I'm following until now is that if a component is able to change state without any interaction from other components, then it should have state. As you said an example is the input... at the end we are keeping an input tag that has a value already and is synced with the state... It may be pointless to keep the state synchronized with the input. On the other hand, you don't need to access the DOM when you want to retrieve the value but it's still semantic markup.

Another example might be the `Drawer`. Since you can close it by clicking in the overlay, it has a state and exposes methods to show and hide. But the responsibility of showing also relays in an external component when it's used so... what's best? having a handler for the overlay and an `active` prop? Or syncing methods with the external component interaction?

I'm more with stateless components but I know sometimes we will have to keep state. I guess the usage will tell us what's the best choice but probably I would migrate to stateless components some like the `drawer`, `dialog`.. etc. Feedback is key here, I'm exploring a lot the usage creating the documentation.

soyjavi:
I'm using a preliminary version of react-toolbox for build a new console for the company I'm working now. I know the *big* problem about `the source of true`, but in my experience some components needs delegate the responsibility to their child components. I mean, `drawer` is a great example... imagine if you have a `ParentComponent` who holds 5 `drawers` and each drawer has async methods (for example for GET/POST information from a service). In my opinion is better decouple responsibility in `ParentComponent` because in fact is not its responsiblity. If not ParentComponent holds business code *out of responsibility*. We resolve one problem but we create new one ;) 

I understand that community is trying to move to *stateless* components, but for now I haven't found a good pattern for resolve this.

ustccjw:
Yes, excessive pursuit of stateless component actually make parent and child  components coupling. We should find balance.

soyjavi:
That's the point... Balance between "trending-topics" and "responsible code" ;)

ustccjw:
@javivelasco I think  this PR can merge now.

server.js
```diff
@@ -1,5 +1,3 @@
-process.env.UV_THREADPOOL_SIZE = 100;
```
javivelasco: Why is this removed??

ustccjw: In spec server, it is unnecessary. But I have added it.
Is it  the node-sass issue?

ustccjw: I see https://github.com/jtangelder/sass-loader/tree/fix/threadpool-issue. Maybe it will be fixed.

javivelasco: That's exactly the issue. Probably that branch fixes it, I think as soon as it's released we can delete the explicit setting.

PR #28, opened 2015-11-02
update readme

ustccjw:
Now package main is `./lib`, and lib is only use babel to handle components. sass files only copy to  `./lib`.

Because many apps will ignore 'node_modules' when using babel-loader. For customization(such as `React Toolbox Loader` ), we will retain sass files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ const CustomButton = () => (
 export default CustomButton;
 ```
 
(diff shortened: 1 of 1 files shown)


soyjavi:
+1 (we need +2 for accept this pull request)

javivelasco:
+1

PR #34, opened 2015-11-03
Include pure implementations of each stateful component

danmartinez101:
I'm super excited to have come across this project on echojs.com! I've been toying with creating something very similar because the existing react implementations of material design aren't quite right for my use cases. This project almost perfectly matches though! so yay!

The one thing that would make it perfect for me would be if all components could be entirely driven by props. An example of this (and where material-ui didn't quite work) is the Dialog component. Where the open/closed state is store internally and changed via a function call. There are ways to work around this but it generally results in a lot of extra code wrapping each component and becomes hard to maintain. 

I'm using https://github.com/rackt/redux and friends and really trying to get to the level of purity where the entire rendered application is a reflection purely of state stored in the Redux store.

This PR contains an example of exporting both a stateful Checkbox and a PureCheckbox. The default is still the stateful so most users will get a checkbox that functions "normally". But for users such as myself that are looking to store state entirely external to the react components, the pure option is available.

This is just an example of a single component in order to get your feedback and see if this is an approach that you would be willing/interested in going with.

Awesome work!

### DIFF
--- a/components/checkbox/index.jsx
+++ b/components/checkbox/index.jsx
@@ -4,6 +4,41 @@ import events from '../utils/events';
 import style from './style';
 
 class Checkbox extends React.Component {
+
+  static defaultProps = {
+    checked: false
+  }
+
+  state = {
+    checked: this.props.checked
+  }
+
+  onChange = (event) => {
+    this.setState({checked: !this.state.checked}, () => {
+      if (this.props.onChange) this.props.onChange(event, this);
+    });
+  };
+
+  render () {
+    return <PureCheckbox
+      {...this.props}
+      onChange={this.onChange.bind(this)}
+      checked={this.state.checked}
+      />;
+  }
+
+  getValue () {
+    return this.state.checked;
+  }
+
+  setValue (value) {
+    this.setState({checked: value});
+  }
+
+}
+
+export
+class PureCheckbox extends React.Component {
   static propTypes = {
     checked: React.PropTypes.bool,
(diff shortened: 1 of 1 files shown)


javivelasco:
I love the idea of having all the components stateless. It's also a good idea to have both but it's going to make the project bigger being actually a quite easy implementation (I mean adding the state control). More feedback on this is very welcome but probably I'm for for stateless components as much as possible. That implies removing it from Dialog, Drawer among others (and update dependencies). Thanks for the effort, let's see the approach we are going to take before merging!

soyjavi:
I think we need something like this:

```
import Checkbox from 'react-toolbox/checkbox';
import CheckboxPure from 'react-toolbox/checkbox/pure';
```

janesconference:
+1

danmartinez101:
Cool, thanks for your feedback. Tonight I will start going through the other components and splitting them into Component and ComponentPure files.

javivelasco:
For now we are going to create a new branch to move all components to stateless. Some of them should hold state like the date picker... but state that is hidden to the parents so it's ok. As soon as we have a working build removing as much state as possible we will release an update. It will be quite breaking and will force you to create higher order components for sure. Coming soon

danmartinez101:
Awesome! Thanks :) 

Should I close this PR then?

javivelasco:
Yes, we are going to use your implementation of the pure checkbox but included in this new branch. Thanks!!

PR #57, opened 2015-11-05
<Tooltip>

soyjavi:
Here is a new component, <Tooltip>.A tooltip is useful for show information on hover in any kind of component. <Button> will have a property `tooltip` which instances this component.

```
<div>
    <p>
      Lorem ipsum dolor sit amet, <strong>consectetur<Tooltip label='This is a auto show tooltip' /></strong> adipiscing elit.
    </p>
    
    <Button label='Button with tooltip' kind='raised' accent tooltip='This is a tooltip by property' />
  </div>
```

### DIFF
--- a/components/_mixins.scss
+++ b/components/_mixins.scss
@@ -275,24 +275,3 @@
     }
   }
 }
-
-@mixin tooltip-hover() {
-  @include tooltip-animation()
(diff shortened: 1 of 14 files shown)


javivelasco:
Nice :) 

Feedback about the usage would be very nice here!

PR #77, opened 2015-11-09
Set font URLs as variables

EnzoMartin:
Enables self hosting of the font files if one does not wish to use Google's CDN or doesn't need/want all the font weights

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 lib
 node_modules
 npm-debug.log
+.idea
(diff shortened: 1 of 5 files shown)


lycon2014:
With my WiFi Configuration would be of expounding measure for personal and business that is one the keys to my ambition

lycon2014:
Which is a perfect fit I think

javivelasco:
Yes, we definitely need to expose font imports to variables. I spoke to @EnzoMartin yesterday and I think we need it... but test are not passing in the PR for some reason and also maybe we can pick a more intuitive name like `$default-font-family` or `$default-font-import` or something similar.

But yes, happy to include it sooner than later :)

javivelasco:
Sorry we forgot about the PR. 
Merged, thanks @EnzoMartin!

PR #90, opened 2015-11-11
Fix disabled ripple click error

fmatoss:
If ripple prop is false, this.refs.ripple will be undefined and will throw a TypeError every time handleMouseDown is called.


javivelasco:
Great, thanks!

Issue #99, opened 2015-11-11
"Autocomplete" page's playground will get 'TypeError' using IE 11


queryxl:
[Autocomplete page](http://react-toolbox.com/#/components/autocomplete)
Click on "Load in playground", it will show 
```
TypeError: Object doesn't support property or method 'from'
```
Using IE 11, win10. Edge and Chrome works.

javivelasco:
Ok! Thanks for reporting!

nathanmarks:
@javivelasco FYI you can always use virtualbox + images from http://modern.ie

javivelasco:
Very nice, I'll give it a try. Thanks @nathanmarks !

nathanmarks:
@javivelasco If you have vmfusion or parallels, they tend to perform a bit better than virtualbox for these.

PR #100, opened 2015-11-12
Added babel polyfill to docs to fix ie11 bug

nathanmarks:
ie11 doesn't seem to support [Array.from](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from). I included the babel-polyfill package which fixes this. Ref #99

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -1,3 +1,4 @@
+import './utils/polyfills'; // Import polyfills for IE11
 import App from './app';
 import AppBar from './app_bar';
 import Autocomplete from './autocomplete';
--- a/components/utils/polyfills.js
+++ b/components/utils/polyfills.js
@@ -1,0 +1,5 @@
+import 'core-js/fn/array/from';
+import 'core-js/fn/array/iterator';
+import 'core-js/fn/map';
+import 'core-js/fn/string/starts-with';
+import 'core-js/fn/symbol';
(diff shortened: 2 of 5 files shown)


javivelasco:
How big does the build get including the polyfill? I was using `Object.assign` in a couple places and I needed the polyfill, but being just two use cases I'd rather avoid its usage in order to keep a smaller build. Maybe it could be better to change a bit the implementation?

Thanks for the PR :)

nathanmarks:
I used babel polyfill as a bit of a catch-all to prevent similar situations
arising as its just a docs site.

However, if you prefer I'll replace it with a specific polyfill.

javivelasco:
Ok, being just in the docs it's ok for me, but in the components we should prevent ES6 features cause we cannot foce people to use a polyfill neither ship the components with it included. Apparently it's 56kb which seems too much to use just a couple of function calls.

Lets add it for now to prevent browser crashing and seek for the ES6 features used in components!
Agree?

nathanmarks:
@javivelasco The `Array.from` function is being called by a helper inserted by babel to polyfill an es6 array spread feature

image

javivelasco:
So do you think we should ship the components with the polyfill included?

nathanmarks:
@javivelasco I've gone ahead and adjusted the PR to include the specific required polyfills from https://github.com/zloirock/core-js in the main component library.

With these polyfills in place, the demo now works in IE11. This would be an alternative to loading the entire polyfill library.

javivelasco:
Very nice @nathanmarks !!!
Thanks :+1: :dancer:

PR #109, opened 2015-11-15
don't hide navigation while playground is loaded if screen is 1200px+

liesislukas:
don't hide navigation while playground is loaded if screen is 1200px+ in width

Also added .idea folder to git ignore. Just don't track project files from popular IDE PhpStorm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 lib
 node_modules
 npm-debug.log
+.idea
--- a/docs/app/components/layout/main/index.jsx
+++ b/docs/app/components/layout/main/index.jsx
@@ -49,6 +49,7 @@ class Main extends React.Component {
     const examples = document.getElementsByClassName(this.LOAD_EXAMPLE_CLASS);
     Array.prototype.forEach.call(examples, (exampleNode, idx) => {
       const exampleCode = components[this.props.params.component].examples[idx];
+      this.refs.playground.loadCode(exampleCode);
       ReactDOM.render(
         <LoadExampleButton onClick={this.handlePlaygroundLoad.bind(this, exampleCode)} />,
         exampleNode
--- a/docs/app/components/layout/main/style.scss
+++ b/docs/app/components/layout/main/style.scss
@@ -70,6 +70,38 @@
   }
 }
 
+/*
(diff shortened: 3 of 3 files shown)


javivelasco:
I'm ok with the `.idea` folder and the query for RWD but not sure about loading each example without clicking. Personally I think people might want to keep the state of the playground while they are navigating through the documentation and with approach it's not possible. Don't you find it useful?

liesislukas:
Before, it was impossible to save previous playground's code, because if you navigate to another component - you hide playground. So now it's just loading same thing with less clicks. 

I found it useful to see active code instantly, because imagine first time visitor, as I was few days ago :) Looking through the list of components and just want to see those in action. If code is auto-loaded without any clicks, you get the result asap. I think it would be very useful to have single page just with all components visible, to get idea of what you may get if you use this lib.

If visitor is not first time. If you navigate to some menu, you expect to see content of that menu item, not previous.  

I think if user wants to keep state of playground could just open new tab/window. It's natural way for me. If you think, autoloading should be optional, there could be a checkbox for this near the "Load in playground" button.

javivelasco:
I see, probably it's more a UX issue. The playground can be shown at any time with the right up corner button. You can close it and open and if you don't load a new example on it, it will keep your changes. That allows you to load say a `Dialog` example and then navigate to the `TimePicker` to see the syntax and add it to the previous example. With this change you propose that's not possible.

Anyway, I totally get your point. It could be very easy to set a `Switch` on appbar or other visible place so you can decide if the examples should be **autoloaded** or not. How does it sound??

Thanks for the PR :)

liesislukas:
I think switch would be a solution. 

What's next is to think about default. ;) I suggest to add Google analytics tracking of that action and after some time you will see which default should be. It's hard to tell without data.

https://developers.google.com/analytics/devguides/collection/analyticsjs/events

javivelasco:
I think it's a great idea. If you want to remove the forced code loading I can merge this PR and open an issue to add a switch or something to toggle autoloading!

liesislukas:
I'm little out of time now, but i think that should be done :+1:

javivelasco:
I'll do it later and merge after that

soyjavi:
+1 (We need +2.. c'mon  @javivelasco )

PR #119, opened 2015-11-17
Use cross-env to set UV_THREADPOOL_SIZE. Fix #83

### LINKED ISSUE #83: [docs] npm start hangs forever in webpack build [Windows]
Repo has some issues with wepack alias + sass loader on Windows machines at least.

See following sass-loader issues:

https://github.com/jtangelder/sass-loader/issues/149

https://github.com/jtangelder/sass-loader/issues/173

When running npm run build, I get file unreadable errors because of the alias not working correctly in sass-loader.

When running npm run start, the bundler never finishes. I can only get npm run start to work, if I remove alias in webpack. But then, obviously, the alias imports in sass don't work.

Just bringing awareness that There is a broken dependency currently in the project.

NogsMPLS:
Weirdly enough, Fixes #83

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -4,7 +4,7 @@
   "description": "Documentation for React Toolbox",
   "author": "React Toolbox Team (http://github.com/react-toolbox)",
   "scripts": {
-    "start": "node ./server",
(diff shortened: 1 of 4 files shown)


javivelasco:
Thanks!

PR #125, opened 2015-11-18
Proposal to change class list generation in components

jonathanp:
What do you guys think about using something like this for generating classname lists in components?

I think it's cleaner as it's more declarative and gets rid of mutation and repeatable code.

If you think it's a good idea I'd be happy to implement it in the rest of the components.


javivelasco:
Hi @jonathanp !!

Yes, I like the proposal and we are probably moving to it soon. @nathanmarks is already using it in the cards branch I think. When it get merged we can start working on it. This afternoon I'm going to try your PR and probably merge.

Can you update it with the latest changes from master?
Thanks! :)

jonathanp:
Np, it's updated now.

Aha, I didn't see @nathanmarks' PR before I created mine. We're using different packages for creating the class name string though, [`classnames`](https://github.com/JedWatson/classnames) vs [`classnames-minimal`](https://github.com/jonathanp/classnames-minimal) :)

nathanmarks:
@jonathanp I need `classnames` as I use `classnames/bind`. Can we stick with the proper repo instead of your fork?

jonathanp:
@nathanmarks: with ES6's computed property names syntax you really don't need the `bind` function though. You can write it with `classnames-minimal` easily like this...

```javascript
import classNames from 'classnames-minimal';
import styles from './style.css';

const classes = classNames({
  [styles.root]: true,
  [styles.disabled]: this.props.disabled,
  [styles.highlighted]: this.props.highlighted
});

console.log(classes); // => contains the class names generated by CSS modules and filtered by our predicates
```

Sure, you can use the `classname` package in the same way but if you only need a small subset of the API I think the 5 LOC version should be preferred over the 173 LOC one.

nathanmarks:
@jonathanp 

Nice tip about the computed property names. I just took another look though and there's a couple of other nice features from classnames missing, including the ability to pass in any number of arguments which is a commonly used feature.

Most people are used to using the API of https://github.com/JedWatson/classnames and you can still do what you want to do with it. 173 LOC is really not a big deal...

nathanmarks:
@jonathanp 

Big issue I just encountered while testing it -- this restricts what users can pass into the components via the `className` attribute. I just tried swapping this into my own app on a couple of components and had issues with passing through the `className` which may not be an object.

jonathanp:
@nathanmarks: That shouldn't be a problem. Since [`this.props.className` is a string](https://github.com/jonathanp/react-toolbox/blob/refactor-classname-generation/components/input/index.jsx#L9) the class name passed down as a prop will be included in the generated string [because of this line](https://github.com/react-toolbox/react-toolbox/pull/125/files#diff-18c5980c2c017c14bdf52f5ee38d0680R60).

I don't see any situation where this API won't be sufficient if you have access to ES2015 syntax.

nathanmarks:
@jonathanp If you are intent on pushing this new implementation, can you at least add the ability to pass in N arguments?

This isn't about your API not working, it's about familiarity with an existing API from the go-to for this functionality in terms of 3rd party packages (500k NPM downloads in the last month).

jonathanp:
@nathanmarks If you could provide a valid reason for extending the API I would be happy to consider it, but I've been trying to demonstrate with my code examples that there really isn't any reason why this is needed. That's the whole point of the `classname-minimal` package, to be a more slim version of `classname`.

@javivelasco what's your take on this?

nathanmarks:
@jonathanp It's not a big deal, I prefer to stick to a familiar API -- and I think we tend to have different styles of using the original package. Here's a snippet from my code:

```jsx
const {
  accent,
  children,
  className,
  icon,
  primary,
  image,
  ...otherProps
} = this.props;

const classes = classNames('avatar', { accent, primary }, className);

// My use explained
// classNames(class, { optional classes }, user classes)

```

I just find it easier to identify what's what at a glance. As far as I know though I could be misusing it!

nathanmarks:
@jonathanp I think both options have pros and cons -- i'm not that fussed and it's not my decision anyways :)

javivelasco:
I like both approaches but I would go for the most used implementation since it would be easier for contributors to add new code and debug errors. @jonathanp I understand your implementation and why did you did it, but I'd like to stick to the most used option. Having a bigger set of features is not a big deal in this scenario IMHO even though they're not all used.

Let's move the current implementation and see how it looks like, we are always free to change again :)

jonathanp:
I understand, np :) I've updated the PR to use the `classnames` package!

javivelasco:
Merged!

FYI, as soon as @nathanmarks finishes the PR for the new `Card` component we are going to migrate all `className` definitions to use `classnames` package in a new branch. @jonathanp feel free to give feedback about the final approach we're gonna take!

Thanks! :)

PR #135, opened 2015-11-19
[autocomplete] Fix #134

### LINKED ISSUE #134: [autocomplete] Impossible to type in the input
# Autocomplete

The input does not behave as expected since it is impossible to type some text.

Bernardstanislas:
# Autocomplete

## Bug

Impossible to input some text.

## Patch

The `state` attribute set in the change handler was wrong, had to change it from `event` to `query`.


javivelasco:
Thanks, can't even understand how we missed it :grimacing:

PR #153, opened 2015-11-22
Removed font icon style dependency, fixes #151

### LINKED ISSUE #151: Icon font should not be dependent on updating this package.
I think this is a pretty important dependency to remove.

I was trying to use a couple of icons just now from the font (via the button component) and they wouldn't work. I realized that you guys are currently defining all of the icon classes manually for the icon props to map to.

This isn't a good idea IMO as it means that whenever google change their icon font, this package is immediately outdated without the user even knowing as it is even calling the google font.

nathanmarks:
No extra stylesheet necessary.

This means users will have to remove the dashes and use underscores instead though like the official font.

Or we can include a search/replace to swap out any dashes for underscores, but it'd be nice to keep things clean and not have that :)


nathanmarks:
Weird that it is failing, `npm run lint` [works perfectly](https://dl.dropboxusercontent.com/s/plenen2rsy00q52/2015-11-22%20at%2010.09%20AM%20%281%29.png) on my system and the error in that check is does not make sense when you look at the source code.

BTW @javivelasco, I swapped out `<i>` for `<span>` as it is more semantic. `<i>` is often used with icon fonts for brevity, not for semantic correctness.

PR #165, opened 2015-11-24
changing months in datepicker (Firefox) failing

andmilj:
in Firefox I am not able to change months in date_picker component.

Firefox is complaining that event object is undefined. 

Event is not sent to handler functions incrementViewMonth and decrementViewMonth (Calendar component)


javivelasco:
Missed that, thanks!

PR #187, opened 2015-12-03
Fix ripple positioning when page is scrolled

jeffreyjw:
When the page is scrolled, the ripple position gets off and is not visible after scrolling few pixels. This PR fixes that behavior, so that the ripple stays where it should


javivelasco:
I didn't notice since usually I wrap the app with `App` component :)
Thanks for the PR @jeffreyjw!

PR #200, opened 2015-12-07
Prevent avatar from getting squeezed

aeickhoff:
If caption or legend of a list item with an avatar are very long, the avatar got squeezed, because it was allowed to shrink in its flex container.


javivelasco:
Thanks!

javivelasco:
Thanks!

PR #206, opened 2015-12-12
Range option added to Slider

nocon:
I've added a Range option to Toolbox Slider:

screen shot 2015-12-12 at 18 07 16

It's not polished/refactored yet, but I could use a review before I finish it. Thanks!

### DIFF
--- a/components/progress_bar/ProgressBar.jsx
+++ b/components/progress_bar/ProgressBar.jsx
@@ -54,10 +54,6 @@ class ProgressBar extends React.Component {
       return {};
     }
   }
-  
-  rangeStyle () {
-    
-  }
 
   renderCircular () {
     return (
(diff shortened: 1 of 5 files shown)


javivelasco:
I think this is useful but I don't know if it's documented in Material Design spec.
Could you make Travis pass and check if they are officially documented?

javivelasco:
Ping here!

nocon:
Range slider is not documented in Material Design spec. Do you still want me to make it pass Travis checks, or should this be shelved until it becomes part of the spec?

javivelasco:
I would rather keep it for now if it's not part of the spec. It would be nice to develop it in another package so you can include useful components modifications that are not part of the spec optionally

scrawfor:
I have a need for something like this. Did it, or something like it ever get added to another package anywhere?

javivelasco:
I will try to make a demo on how can you compose your own with Slider subcomponents but I need some time to make it work!

KeizerDev:
Any update?

tonymaibox:
Would like to know if there's an update also! Many thanks in advance 😄

PR #210, opened 2015-12-15
handle mouse leave in IconButton like Button does, fixes #208

### LINKED ISSUE #208: IconButton doesn't have onMouseLeave handled like Button does

jackielii:


javivelasco:
Thanks!

PR #212, opened 2015-12-15
Allow CardMedia to have flexible height per documentation

pdf:
The documentation for CardMedia specifies that if the `aspectRatio`
property is not set, the media will have a flexible height.  The
existing implementation did not provide this, and would instead always
render with a height of `0`, effectively hiding the element.

This change should bring the component into line with the documentation.


javivelasco:
Thanks!!

PR #217, opened 2015-12-16
Add maxLength attribute to input component

avijohnson:
@javivelasco The maxLength attribute was not working, and it was because it was being displayed but not being added to the element.


javivelasco:
Great, thanks for the PR! :)

PR #223, opened 2015-12-18
Replace `overflow-y: scroll` with `overflow-y: auto`

### LINKED ISSUE #218: `overflow-y: scroll` vs `overflow-y: auto`
A number of components use `overflow-y: scroll`, which forces scrollbars to be always visible. I presume to work around this, `commons.scss` hides webkit scrollbars globally.  There are two problems with this: on non-webkit browsers, scrollbars are visible even when they are unnecessary; and scrollbars are hidden in webkit browsers, even on elements where they may be desirable.

I propose using `overflow-y: auto; -webkit-overflow-scrolling: touch;` in all components currently using `overflow-y: scroll;`, and if maintainers feel strongly about hiding scrollbars on specific components, those can individually specify the relevant webkit properties.

Happy to send a PR for the former if this is acceptable, of for both if maints can specify the components that require scrollbars hidden (though I'm not convinced this is a great idea).

pdf:
A number of components used `overflow-y`: scroll, which forces
scrollbars to be always visible. To work around this, `commons.scss` hid
webkit scrollbars globally. There are two problems with this: on
non-webkit browsers, scrollbars are visible even when they are
unnecessary; and scrollbars are hidden in webkit browsers, even on
elements where they may be desirable.

Instead, we use `overflow-y: auto`, and hide webkit scrollbars on
specific components when necessary.  Unfortunately, this requires using
element selectors, but this is the trade-off for better supporting
non-webkit browsers and allowing scrollbars on regular elements (like
body).

Fixes #218

### DIFF
--- a/components/date_picker/style.calendar.scss
+++ b/components/date_picker/style.calendar.scss
@@ -8,6 +8,7 @@
   line-height: $calendar-row-height;
   text-align: center;
   background: $calendar-primary-contrast-color;
+  overflow-x: hidden;
   .prev, .next {
     position: absolute;
     top: 0;
@@ -32,7 +33,8 @@
 
 .years {
   height: 100%;
-  overflow: scroll;
+  overflow-x: hidden;
+  overflow-y: auto;
   font-size: $font-size-big;
   > li {
     line-height: 2.4;
--- a/components/dropdown/style.scss
+++ b/components/dropdown/style.scss
@@ -42,6 +42,10 @@
     top: auto;
     bottom: 0;
   }
+  ul.values::-webkit-scrollbar {
+    width: 0;
+    height: 0;
+  }
 }
 
 .label {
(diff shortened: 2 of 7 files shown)


pdf:
I've hidden scrollbars for webkit on the autocomplete and dropdown lists to avoid flashing scrollbars on transition, otherwise I've left them in place.

components/dropdown/style.scss
```diff
@@ -42,6 +42,10 @@
     top: auto;
     bottom: 0;
   }
+  ul.values::-webkit-scrollbar {
```
javivelasco: We should always avoid tagnames selector

components/date_picker/style.calendar.scss
```diff
@@ -32,7 +33,8 @@
 
 .years {
   height: 100%;
-  overflow: scroll;
+  overflow-x: hidden;
```
javivelasco: Avoid when possible

javivelasco:
It is merged but fucked up the commit message so it's not properly recognized in Github I think.
Thanks!

Issue #92, opened 2015-11-11
Allow passing components to leftIcon and rightIcon on ListItem


adamscybot:
This may well be a recurring theme throughout the lib -- but I have a few use cases whereby I want to pass in more than just a string to an attribute. For example, I have a `<ListItem>` where on the right hand side, I want a custom `<img>` tag rather than an icon. I often want something completely custom their as well.

In https://github.com/callemall/material-ui it is possible to simply pass this straight into the prop. I don't particularly like this, but it does solve the use case.

Another option I can think of would be to allow children on this component which would completely override the "utility" attributes `caption`,  `legend`, `leftIcon` and `rightIcon` and just allow you to put anything in the box.

Thinking about it, I think support for both methods might be appropriate? That would allow these components to be as flexible as you want them to be.

javivelasco:
You are right, we need to be able to give arbitrary content to right and left side. For this first version we are not supporting them but we need it though. I think the best approach would be to have special children overriding the props but I'm not sure how clean can look the implementation. Probably the best is to try both methods and see how does it look both for the usage and implementation.

I'm going to label this issue as an enhancement to work on after we finish the current tasks (move to stateless components like button, dialog, drawer, etc). This feedback is great because I know we are missing some props and use cases for components. For example, almost none of them support `onMouseEnter` and similar callbacks.

Thanks! :smile:

adamscybot:
Its worth noting here that ```caption``` and ```legend``` already support this (albeit with a warning due to incorrect proptypes).

Thanks for the work on this plugin! Enjoying using it.

javivelasco:
Thanks @adamscybot :)

neverfox:
Definitely looking forward to this, as it's currently a dealbreaker for making the leap from material-ui. One of the clearest cases is being able to use the `Avatar` component as the `ListItem`'s avatar.

javivelasco:
This is possible now!

PR #226, opened 2015-12-21
customizable list

kagux:
I've run multiple times into situation where custom behaviour from either element of the `ListItem` was required. Usually it's either of these:
* use button instead icon
* custom components in the middle part
I had to override `renderContent()` in the `ListItem` to use my desired components and import styles from `react-toolbox`. 

This PR:
* backwards compatible ( passing props to configure `ListItem` )
* uses children to set custom components
* allows overriding `ListItem` components with custom ones
* allows mixings props and custom components for partial overriding
  * custom components have precedence
* solves issue https://github.com/react-toolbox/react-toolbox/issues/92

Example:
```
<List>
  <ListItem>
    <ListItemLeft> <span> left </span> </ListItemLeft>
    <ListItemAvatar>
      <img src='https://pbs.twimg.com/profile_images/614407428/s6pTalMzZs-nusCGWqoV.0_400x400.jpeg'/>
    </ListItemAvatar>
    <ListItemMiddle>
      <ListItemCaption> <span> custom caption </span> </ListItemCaption>
      <ListItemLegend> <span> custom legend </span> </ListItemLegend> 
    </ListItemMiddle>
    <ListItemRight> <span> right </span> </ListItemRight>
  </ListItem> 
  <ListItem rightIcon='delete'>
    <ListItemLeft> <span> left </span> </ListItemLeft>
    <ListItemMiddle caption='combine'>
      <ListItemLegend> with props </ListItemLegend>
    </ListItemMiddle>
  </ListItem> 
</List>
```

image

Would love your feedback on this!

### DIFF
--- a/components/list/ListCheckbox.jsx
+++ b/components/list/ListCheckbox.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import ClassNames from 'classnames';
 import Checkbox from '../checkbox';
-import ListItemContent from './ListItemContent';
+import ListItemMiddle from './ListItemMiddle';
 import style from './style';
 
 const ListCheckbox = (props) => {
@@ -16,7 +16,7 @@ const ListCheckbox = (props) => {
         checked={props.checked}
         className={style.checkbox}
         disabled={props.disabled}
-        label={<ListItemContent caption={props.caption} legend={props.legend} />}
+        label={<ListItemMiddle caption={props.caption} legend={props.legend} />}
         name={props.name}
         onBlur={props.onBlur}
         onChange={props.onChange}
--- a/components/list/ListItem.jsx
+++ b/components/list/ListItem.jsx
@@ -7,6 +7,7 @@ import ListItemMiddle from './ListItemMiddle';
 import ListItemLeft from './ListItemLeft'
 import ListItemRight from './ListItemRight'
 import ListItemAvatar from './ListItemAvatar'
+import ReservedChildren from '../reserved_children/ReservedChildren'
 
 class ListItem extends React.Component {
   static propTypes = {
@@ -22,23 +23,6 @@ class ListItem extends React.Component {
     selectable: false
   };
 
-  reservedChildrenTypes = new Set([ListItemMiddle, ListItemLeft, ListItemRight, ListItemAvatar]);
-
-  isReservedChild (child) {
-    return child && this.reservedChildrenTypes.has(child.type);
-  }
-
-  reservedChildren () {
-    let children = {};
-    React.Children.forEach(this.props.children, (child) => {
-      if (this.isReservedChild(child)) {
-        children[child.type] = child;
-      }
-    });
-
-    return children;
-  }
-
   handleClick = (event) => {
     if (this.props.onClick && !this.props.disabled) {
       this.props.onClick(event);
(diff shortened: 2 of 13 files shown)


nathanmarks:
@kagux Great work! I'll let @javivelasco give his take, but a few initial thoughts I feel need addressing:

1. Should use the `react-toolbox/components/avatar/Avatar.jsx` component for avatars in examples
2. Follow material design spec naming conventions for list item text
  - Primary Text
  - Secondary Text
  - [reference from spec](https://dl.dropboxusercontent.com/s/stogp712hz5uv12/2016-01-01%20at%2011.45%20AM.png)
4. I think there are too many components currently that are just `<span></span>` with styling. It adds a lot of noise that is unnecessary. I feel like I should be able to set up a list in this way:

##### Simple list of avatars with title/description

```jsx
<List>
  <ListItem>
    <Avatar image={avatarUrl} />
    <ListItemText
      primary={title}
      secondary={message}
    />
  </ListItem>
</List>
```

##### Clickable list with secondary clickable action

```
<List>
  <ListItem onClick={this.handleItemClick}>
    <Avatar image={avatarUrl} />
    <ListItemText
      primary={title}
      secondary={message}
    />
    <FontIcon onClick={this.handleItemStar}>star</FontIcon>
  </ListItem>
</List>
```

I haven't reviewed the UX interactions yet -- but will do later. @kagux what do you think to the points above?

kagux:
Thanks for taking your time, @nathanmarks !
1. Great point about avatar, totally missed it!
2. and 3. I feel like we have different use cases in mind. As I pointed out in PR description, the point here is to allow putting custom components in place of existing ones, while keeping layout and looks. Real life examples in my case:
* using buttons instead of icons
* inline editing text => have to replace text component completely

For simple use cases like you provided it would be best to stick with default syntax that renders list from props. 

Most components are just spans with styling to provide layout for the list item parts and in simple case just render the text. This could be pulled back to item render. But it's a trade off between item render clean syntax with multiple simple span components or cluttered item render without additional components. I prefer the first solution as it keeps noise out of render and simple components are easy to read.

What do you think?

nathanmarks:
@kagux The problem is the existing `<List>` component doesn't let you do a number of the most prominent examples and uses of the material design list.

A single component with props is not sufficient to pull off the different list variations seen in the spec and google apps.

With my examples above, there is no reason why you couldn't use custom components in there. The alignment is managed by some simple flex properties -- we don't need 3 extra components to allow "custom" components. The 3 components you have for the inner content are essentially alignment helpers that can't be re-used.

What I'm proposing is a solution that gives enough flexibility to reproduce spec examples and uses seen heavily in android apps, combined with the ability to easily drop custom components in wherever you want.

nathanmarks:
@kagux Just to summarize my thoughts:

I think the list component should be customizable in a way that it allows you to easily drop in different components to recreate various different usage scenarios that are visible in both the spec and google apps.

In that system, designed properly, I see no reason why 'custom' components cannot be dropped in too.

kagux:
I see where you're going with this @nathanmarks. It does make sense to simplify API as it has redundant parts.
I gave some thought to the List API over weekend, it should be technically possible to define a list item contents using just basic components. I've got 2 concerns:

* How to figure out what component to use for central text part.
Everything before this component goes to the left side, everything after it to the right.

```
<List>
  <ListItem onClick={this.handleItemClick}>
    <Avatar image={avatarUrl} />
    <ListItemText
      primary={title}
      secondary={message}
    />
    <FontIcon onClick={this.handleItemStar}>star</FontIcon>
  </ListItem>
</List>
```
how would I render input field instead of primary title?
`<ListItemText><MyConponent/></ListItemText>` ?
in angular demo they seem to allow any non special tag to become item message
like
```
<ListItem>
   <Avatar />
   <MyComponent />
   <FontIcon />
</ListItem>
```
 * How to deal with backwards compatibility

javivelasco:
Great work guys! I'm back from holidays and I need to think about this and try some example. I'm a little out of context right now. @kagux IMHO the approach you are taking in the PR feels a bit 'hacky'. It wasn't easy for me to understand the usage of the relation parent/children, it's easier to just compose components the way @nathanmarks says but I'm not sure about the implementation yet. I think this should be very similar to `Cards`. We can agree here how we would want the API to be a create an issue to work on it together.

kagux:
It's been a busy week after holidays. Let's agree on API and I will put together first implementation.

Default list item content with custom actions
```
<List>
  <ListItem onClick={this.handleItemClick}>
    <Avatar image={avatarUrl} />
    ....[left actions]
    <ListItemContent caption={title} legend={message} >
    <FontIcon onClick={this.handleItemStar}>star</FontIcon>
    ....[right action]
  </ListItem>
</List>
```

Custom item content:
```
<List>
  <ListItem onClick={this.handleItemClick}>
    ....[left actions]
    <ListItemContent>
      <MyComponent/>
      .....
    </ListItemContent>
    ....[right actions]
  </ListItem>
</List>
```

For backward compatibility props can be converted to corresponding components and prepended to left or right actions.

kagux:
@javivelasco @nathanmarks what do you think of proposed API?

javivelasco:
Sorry @kagux about taking so long to respond. I think the API you proposed is neat.
What about @nathanmarks s feedback on this?

Issue #230, opened 2015-12-23
Replace transition-property to transition, because Autoprefixer could not support any cases of transition-property and other transition-*


itsNikolay:
`$ webpack -p`

/node_modules/react-toolbox/lib/animations/zoom-in.scss:1:141: Replace transition-property to transition, because Autoprefixer could not support any cases of transition-property and other transition-*

Nodejs 4.3.2
"react-toolbox": "^0.14.0"
"autoprefixer": "^6.1.2",
"autoprefixer-core": "^5.1.0",

Bartekus:
I've submitted a new pull request which will take care of all these errors as well as bring the entire development up to date, dependencies wise.

javivelasco:
Great, I'll check the PR! Thanks!

neverfox:
Would love to get rid of these warnings. When is the next release planned?

javivelasco:
Today!

shidhincr:
Yay!

javivelasco:
Alredy done guys. Probably another patch will come soon since people are contributing with small fixes constantly. And I love it

carlokid:
Is this fixed?

Issue #59, opened 2015-11-05
.babelrc incompatible with Babel 6


joshuaegclark:
the current .babelrc uses the "stage: 2" option which is no longer supported in the latest Babel, which causes it to throw an error if you are using Babel >= 6.0.0.

javivelasco:
Ok! Updating config to babel 6. Thanks for reporting!

javivelasco:
I will be happy to accept any PR with a HMR configuration using Babel 6!

hasyee:
What's about Babel 6 updating? Thanks!

javivelasco:
Yep, as I said in the previous comment, contribution here is very welcome. BTW, I don't know yet about babel 6 support in HMR which is very useful for development...

nathanmarks:
Waiting on this, https://github.com/gaearon/babel-plugin-react-transform/issues/46

javivelasco:
Perfect, so as soon as we get support we can update! 
Thanks @nathanmarks !

Bernardstanislas:
@javivelasco I think you can use [react-hot-loader](https://github.com/gaearon/react-hot-loader) to achieve HMR with Babel 6, we did this in [our projects](https://github.com/KleeGroup/focus-comments/blob/develop/webpack.config.js#L34) and it works like a charm :)

javivelasco:
Awesome! Gonna work on it during the afternoon. Thanks @Bernardstanislas :)

carloscuatin:
@javivelasco create a roadmap for upgrade to babel 6

javivelasco:
Yeah we should. I was waiting for HMR to be updated and it seems it's coming. I'll have some time to work on toolbox during the weekend and I'll try to release some changes and set up next priorities. This should be one of them

carloscuatin:
@javivelasco create a roadmap, i create pull request for upgrade

guerrero:
@javivelasco It seems that this issue can be closed without further actions

javivelasco:
Yep! thanks @guerrero !

PR #233, opened 2015-12-23
Issue #59 - Babel 6 Upgrade

Bartekus:
As per issue: .babelrc incompatible with Babel 6 #59

Bebel 6 has been integrated, Stage-0 has been added, React and Webpack have also been updated.
In addition webpack configuration files have been slightly tweaked and adjusted to babel 6 changes. 

Currently the Autocomplete.jsx file is disabled due to bug in babel https://phabricator.babeljs.io/T6712

Aside from toolbox-loader, and excluding highlight & sinon, everything else has been updated.

With all test passing this pull is now complete.


javivelasco:
This is great. The only thing that I don't like is the need to disabled `Autocomplete` component. I think it should be fixed before merging so we can have the `dev` branch working nicely and ready to merge at any point. What do you think? It's going to take too long, is there anything that can be changed to avoid the bug?

Thanks @Bartekus great work!

Bartekus:
I'm looking into fixing Autocomplete but after taking two days off I'm only getting back into it now so a little bit more time is needed before I come up with a solution. Thank you for patience and stay tuned!

carloscuatin:
what is the problem in autocomplete ??

Bartekus:
The problem is this:
```
SyntaxError: components/autocomplete/Autocomplete.jsx: "suggestions2" is read-only (This is an error on an internal node. Probably an internal error. Location has been estimated.)
```
and it relates to this: https://phabricator.babeljs.io/T6712

Bartekus:
Unfortunately at the moment I'm lacking skills to fix this problem properly and hacking it into passing a test won't guarantee artifact-free operation hence I'd like to sway away from that option. 
I will monitor the open tickets for the solution on the babel side and implement the solution as soon as one is found.

javivelasco:
Great, don't worry! :)

vicentedealencar:
looking at Autocomplete.jsx I see
```
suggestions () {
  const suggestions = new Map();
```

maybe this is the source of error related to `constant inside class method of same name`? https://phabricator.babeljs.io/T6712

maybe changing this const name fix it

I wasn't able to build it on windows, so I couldn't try it

langri-sha:
Keeping the books in order: this also fixes #230.

javivelasco:
Fixed and merged! Thanks everybody! :D

Issue #228, opened 2015-12-22
Ripple decorator breaks extensibility of components


kagux:
In version 0.13.4 I could easily extend components to override partial rendering methods. With latest release of 0.14 I can't do it anymore. Components that are decorated with Ripple can no longer be extended.

```
import React from 'react';
import Ripple from '../../components/ripple';


export default class RippleTest extends React.Component {
  render () {
    return (
      <section>
        <h5>Extending Rippled Component</h5>
        <div> Parent: <Parent/> </div>
        <div> Child: <Child/> </div>
        <div> Child of rippled parent: <RippledChild/> </div>
      </section>
    );
  }
}


class Parent extends React.Component{
  whoami () {
    return 'I am a parent';
  }

  render () {
    return <b>{this.whoami()}</b>;
  }
}

const RippledParent = Ripple()(Parent);

class RippledChild extends RippledParent {
  whoami () {
    return 'I am a rippled child';
  }
}

class Child extends Parent {
  whoami () {
    return 'I am a child';
  }
}
```

![image](https://cloud.githubusercontent.com/assets/1159573/11956575/9173170c-a8bb-11e5-9c1f-5e0413c566ee.png)

javivelasco:
@kagux I see your point. I think I mentioned decorator while I wanted to say higher order component. For your use case it makes sense to extend the component decorating it more than wrapping it with a HOC. I need to test some use cases and think about this. Feedback here appreciated.

Also, I still need to check your PR which I see goes in the same direction. I've never needed to extend components before so I don't know if it's a common approach and need to dig deeper on it. Be patient, I'm not having much time these days. BTW, thanks for the PR!

kagux:
@javivelasco I'm of two opinions here:

* On one hand component's object methods are sort of private API and should not be depended upon. The correct path for extension is letting end users pass their own components/callbacks via public API (props and children).

* On the other hand it gives developers a way to quickly bend the component to their will while a better solution is being discussed/developed/reviewed in PR. In fact clean classes with nicely broken down methods was one of the key points why we went with your library. It is great in other aspects as well!

In this particular case, what would you say if Ripple injected itself in component's children instead of wrapping? I'd be willing to submit PR with POC for this.

No worries about my PR. It's Christmas time, best spent with family and friends rather than PR reviews! :)

PR #235, opened 2015-12-26
inject ripple into component instead of wrapping it

kagux:
POC implementation as follow up to https://github.com/react-toolbox/react-toolbox/issues/228 discussion.

Instead of using HOC for ripple, this patch injects it into children, leaving component class available for extensions.

### DIFF
--- a/components/ripple/AddRippleToComponent.jsx
+++ b/components/ripple/AddRippleToComponent.jsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import RippleComponent from './RippleComponent';
+
+const AddRippleToComponent = (Component, {centered, className, spread}) => {
+  return class RippledComponent extends Component {
+    static propTypes = {
+      ...Component.propTypes,
+      disabled: React.PropTypes.bool,
+      ripple: React.PropTypes.bool,
+      rippleCentered: React.PropTypes.bool,
+      rippleClassName: React.PropTypes.string,
+      rippleSpread: React.PropTypes.number
+    }
+
+    static defaultProps = {
+      ...Component.defaultProps,
+      ripple: true,
+      rippleCentered: centered,
+      rippleClassName: className,
+      rippleSpread: spread
+    }
+
+    _rippleHandleMouseDown = (event) => {
+      if (!this.props.disabled) {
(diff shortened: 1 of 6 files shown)


kagux:
@javivelasco I know you're busy lately. I'd appreciate if you could find a moment to take a look at this PR.

javivelasco:
Yeah sorry. I would ask @nathanmarks to took a look on it too. I want to download the code and try some stuff but truth is it looks a little hacky for me. I see more or less the same approach you proposed [here](https://github.com/react-toolbox/react-toolbox/pull/226#issuecomment-172213425) and it makes me to be a little lost in the code.

I would like to have more feedback about this from the community. Maybe I should just update myself, anyway it's great work from your side. Let's try to get some more feedback!

nathanmarks:
@javivelasco The code makes sense to me. It's a clever solution to the problem IMO. Personally, I don't like extending components -- but people should be able to if that's their cup of tea.

@kagux Quick question though -- if you're already customizing the components to this degree, how does this sound as an alternative solution:

On the dev branch right now, I just took a peek and when it hits master, you can just import the named export from the component (https://github.com/react-toolbox/react-toolbox/blob/dev/components/button/Button.jsx#L82) and then you can extend to your heart's desire ripple-free and also apply the ripple if you want it.

Obviously, the PR is a nice solution to fix that out of the box extensibility with the ripple, but on the other hand it does increase the complexity a little and adds more LoC. Like you previously mentioned @kagux , people generally treat library component APIs that are undocumented (or not advertised as extensible) as private. 

What do you think to the solution of extending the named export directly from the Component to avoid any HoCs applied by the library all together?

kagux:
@nathanmarks I came here to say pretty much the same when I saw your reply. With this PR I wanted to explore a general solution. But if the added complexity is not justified by odd use cases, then exporting named unrippled components would be my second choice. 

What would be a good naming convention in such case? Would exporting `RippledButton` by default and named `Button` be confusing?

javivelasco:
@kagux if the alternative fits for you, I'd rather keep the code easier. If not then we can just merge but I'd like to have simple code. Maybe the button without ripple can be called something like `RawButton`? Please, don't get me wrong. It's great work and a very wise solution and I really appreciate your PR :smile:

PR #245, opened 2016-01-05
Tooltips works with Link components

karlmacklin:
#### Change
I've changed the Link component to render children. I've updated the docs example and component readme to reflect this as well.

#### Note!
* Please review the changes to Link.jsx to see if it's a reasonable way to make the change. I'm new to ES6, not very well versed with how Tooltip renders stuff and also just assumed it would be best to put the children rendering last in the anchor tag. Have some experienced eyes glance over it.
* To make the tooltip visible I removed the `overflow: hidden;` from Link style. Not sure if it's needed fro something else.

Otherwise, seems good,

### DIFF
--- a/components/link/Link.jsx
+++ b/components/link/Link.jsx
@@ -3,7 +3,7 @@ import ClassNames from 'classnames';
 import style from './style';
 import FontIcon from '../font_icon';
 
-const Link = (props) => {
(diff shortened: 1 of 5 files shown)


javivelasco:
Nice, merged! Thanks!

Issue #254, opened 2016-01-08
CardTitle avatar doesn't have a circular shape if the source image is not square


YoYuUm:
CardTitle avatar doesn't have a circular shape if the source image is not square. This could be solved rendering the image as background-image instead of inside an img element.

javivelasco:
Just answered to this in the PR :)

PR #256, opened 2016-01-08
Add support for non-rounded images as avatar

YoYuUm:
Proposed solution for issue #254


javivelasco:
Thanks for your PR @YoYuUm. I don't think this is the best solution actually. I agreed with @nathanmarks that if you are going to add an image as an avatar it should have the proper cropping. Setting the image as a background has some caveats... an example would be accessibility.

What do you think?

nathanmarks:
@javivelasco @YoYuUm You can pass a custom component to the `image` prop instead of an image url string

javivelasco:
:+1:

PR #261, opened 2016-01-13
Add a way to detect problems before publish

Kikobeats:
I create a more safe process to create a new tagged version.

Basically runs [git-dirty](https://github.com/Kikobeats/git-dirty) to check that your local version is sync with remote.

Also I moved some bumped command into npm scripts to be more easy to edit (check push script).

Typically I bind `npm run clean` to `rm -rf node_modules` but you are using the command for other thing so I call rm inline in the bumped command. Consider if is possible merge two actions.


Kikobeats:
I'm going to make another PR

javivelasco:
:+1:

PR #262, opened 2016-01-14
Add TypeScript definition file and typings entry to package.json

### LINKED ISSUE #199: Any chance of TypeScript definitions support?
It became very popular today, and may be already worth to create ones

hsrobflavorus:
As well as a blurb in README.md. This should resolve #199

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,10 @@ If you browse the resulting markup you will see *data attributes* like `data-rol
 }
 ```
 
+## TypeScript
+
+A TypeScript definition file `react-toolbox.d.ts` is available. It is referenced in `package.json` and should be picked up by the TypeScript compiler when importing from the npm package.
+
 ## Authors and Contributors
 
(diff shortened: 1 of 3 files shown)


javivelasco:
Awesome! Thanks!!

xogeny:
:+1:

hsrobflavorus:
@xogeny Thanks for the initial definition layout!!! I couldn't have done this without it.

alltom:
Do you happen to have a public project that uses react-toolbox with TypeScript that I could use as an example? I'm having trouble with the configuration.

javivelasco:
If somebody has one, I think we should share the link in the Readme too!

xogeny:
If I get some time, I might try to incorporate this into [my Typescript+React+Vada implementation of TodoMVC](https://github.com/xogeny/vada-todo).  All UI stuff is represented in this [single file](https://github.com/xogeny/vada-todo/blob/master/src/view.tsx).  Of course, I'd gladly accept a pull request for an alternate implementation from somebody else too.

hsrobflavorus:
@alltom Please check out the updated definition file that was added in #293  

There were some problems (and possibly slight changes on the react-toolbox side) with the previous version, but the new one is merged in and should hopefully fix the issues you were having.

PR #265, opened 2016-01-16
Update package.json

trungtin:
sass command not run on Window with single quote, haven't test on others.


javivelasco:
I don't know how should this run on Windows but it's a script for building purposes. The thing is that test are broken because of the change you proposed. Because of that I can't accept it. Maybe you'd want to try other way in Windows to make the test pass too?

PR #267, opened 2016-01-19
Enable any Component to behave as Icon

epilgrim:
Hey there! first of all, thanks for the awesome project you have created. Really enjoying working with it.

I'm wondering what do you think about the following usecase. Sometimes I need icons (social icons for example) that are not in the material-icons set. In those cases, I have to replicate the icon behavior in the outside

eg:
```html
<span>
  <svg.....></svg>
  <Input ..../>
<span>
```

and then of course style that to behave as a native icon.

What if instead of passing the icon as a name, we enable to pass a full react element:
```html
<Input icon={<svg ...></svg>} ... />
```
This way, it will be natively treated as an icon. For the POC I just loosened the validations, from `string` to `any` in the FontIcon and the Input components. If the idea is accepted, I can extend the behavior, and polish it in the whole library.

### DIFF
--- a/components/font_icon/FontIcon.jsx
+++ b/components/font_icon/FontIcon.jsx
@@ -2,7 +2,10 @@ import React from 'react';
 import ClassNames from 'classnames';
 
 const FontIcon = ({ children, className, value, ...other}) => {
-  const classes = ClassNames('material-icons', className);
+  const classes = ClassNames(
+    {'material-icons': !React.isValidElement(value)},
+    className
+  );
   return (
     <span className={classes} {...other} >
       {value}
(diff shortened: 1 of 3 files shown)


kagux:
that would be a nice feature!
`FontIcon` would be better off validating icon as either `string` or `object`

javivelasco:
Thanks for your kind words and the PR @epilgrim !! 
I think it's a nice feature too @kagux :) 

Merged!

epilgrim:
Thanks @javivelasco for taking a look. 
Although it's merged, I will finish this pull request, loosening the validation in the other components, and making it stricter on the FontIcon (string or object as pointed out by @kagux )

Hopefully today or tomorrow.

javivelasco:
Perfect! Thanks :)

PR #271, opened 2016-01-22
allow passing components to card title and subtitle

kagux:
We use custom components to provide placeholders. But it could also be used to attach events to title/subtitle.


javivelasco:
Great, thanks again @kagux!

PR #272, opened 2016-01-22
export raw components

kagux:
As a follow up to previous discussion in https://github.com/react-toolbox/react-toolbox/pull/235


javivelasco:
Very nice!! Thanks @kagux

PR #274, opened 2016-01-22
Enabled passing components as icons in all the components

epilgrim:
As promised, here are the rest of the changes to make this feature complete.


javivelasco:
Great! Thanks

PR #275, opened 2016-01-22
close dropdown when click event is triggered outside the component

### LINKED ISSUE #244: Dropdown is not closing after clicking outside of it.
Dropdown is closing after select some element from list. It will be perfect, when dropdown also close when user click outside of it.

epilgrim:
fixes #244

### DIFF
--- a/components/dropdown/Dropdown.jsx
+++ b/components/dropdown/Dropdown.jsx
@@ -37,6 +37,12 @@ class Dropdown extends React.Component {
     }
   }
 
+  componentDidUpdate (prevProps, prevState) {
+    if (prevState.active && !this.state.active) {
+      events.removeEventsFromDocument({click: this.handleDocumentClick});
+    }
+  }
+
   handleDocumentClick = (event) => {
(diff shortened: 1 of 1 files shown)


components/dropdown/Dropdown.jsx
```diff
@@ -30,6 +31,18 @@ class Dropdown extends React.Component {
     up: false
   };
 
+  componentWillUpdate (prevState, nextState) {
+    if (!prevState.active && nextState.active) {
+      events.addEventsToDocument({click: this.handleDocumentClick});
```
javivelasco: I think here a remove event listener is missing, isn't it? This would add as many event listeners to the document as times you open the dropdown. In other components we added the event listener at mount and removed at unmount. I know this means having an event listener for each component but definitively it should be removed when there is not needed anymore.

Thanks for the PR, I'll be happy to merge if you add the remove :)

epilgrim: you are totally right. I saw it when following how you did it in the other components, but then forgot completely.

javivelasco:
Great! Thanks for fixing it :)

Issue #264, opened 2016-01-15
Dialog content is rendered on componentDidMount


sars:
Hi there!
I noticed, that dialog content is rendered immediately after the initial rendering occurs.
https://github.com/react-toolbox/react-toolbox/blob/dev/components/overlay/Overlay.jsx#L24

Isn't it nicer to render it of first active=true prop is received?

javivelasco:
Sure, makes a lot of sense. Do you want to PR it?

ro-savage:
This should be fixed in v1.0.0. If its fixed, maybe close the issue? @sars

javivelasco:
Gonna close, I think it should!

PR #279, opened 2016-01-25
render overlay children on demand only

kagux:
Currently overlay and its children are always in the DOM and change opacity/position on activation.
With this PR overlay renders children only when active. I had to use `ReactCSSTransitionGroup` to keep animation as children don't change active state anymore. `ReactCSSTransitionGroup` is set up to use `:global` styles to give children animation hooks. The drawback of using `ReactCSSTransitionGroup` is necessity to keep animation duration in sync between JS and CSS.

Fixes https://github.com/react-toolbox/react-toolbox/issues/264

### DIFF
--- a/components/dialog/Dialog.jsx
+++ b/components/dialog/Dialog.jsx
@@ -15,7 +15,11 @@ const Dialog = (props) => {
   }, props.className);
 
   return (
-    <Overlay active={props.active} onClick={props.onOverlayClick}>
+    <Overlay
+      active={props.active}
+      onClick={props.onOverlayClick}
+      animationDuration={props.animationDuration + props.animationDelay}
+      >
       <div data-react-toolbox='dialog' className={className}>
         <section role='body' className={style.body}>
           {props.title ? <h6 className={style.title}>{props.title}</h6> : null}
@@ -32,6 +36,8 @@ const Dialog = (props) => {
 Dialog.propTypes = {
   actions: React.PropTypes.array,
   active: React.PropTypes.bool,
+  animationDelay: React.PropTypes.number,
+  animationDuration: React.PropTypes.number,
   children: React.PropTypes.node,
   className: React.PropTypes.string,
   onOverlayClick: React.PropTypes.func,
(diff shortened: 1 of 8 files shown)


javivelasco:
I need to have a deeper look at this because I see a couple of problems. First, I saw an `!important` that we need to avoid. Also, I think we should use CSS modules for the transition classnames since the fit very well actually. Just check the transitions for the `DatePicker` component. Also, It's important to be able to change the appearance of the overlay to something different from dark with small opacity.

Another thing is that, from a first check, it looks like we need to repeat code to style the overlay in different components. Let me check it deeper during the weekend.

Thanks for your work @kagux

kagux:
thanks for looking into this PR @javivelasco !
1. I hated the `!important` rule, but I already spent quite some time fighting css modules with sass compilers. I hoped someone could help me out here. But to be fair, it sets opacity of invisible layer to 0, so I very much doubt someone would want it different.
2. I had to use React transitions instead pure CSS transitions because components are rendered only on demand. They render in `active` state, so there is no CSS transition at all.
3. Overlay is styled in one place only, in it's own `style.css`. It does expose global CSS classes that are appended by React addon during transitions so that other components can add animation in sync with overlay. So you see same global classnames, but in each component they are used to add styles/animations to single component only.

javivelasco:
Ok, thanks for clarifying. I'll check it during the weekend :)

kagux:
@javivelasco did you get a chance to look into this PR?

javivelasco:
Sorry dude, I know I'm not spending too much time on this lately but I'm having really busy days. I'll try to check on it as a top priority during the next days. Sorry about that :(

kagux:
np @javivelasco, I understand. I see quite a few issues/PR around these components. Would prefer to avoid conflicts (code divergence or merging issues) in case you accept this PR.

ro-savage:
Would love to see this get checked and merged. I am currently using a custom HoC for handling this.


@javivelasco @kagux  Let me know if I can take a look or be of any help.

javivelasco:
Check latest PR!

javivelasco:
And thank all!

PR #282, opened 2016-01-26
List Item customization

kagux:
As follow up to https://github.com/react-toolbox/react-toolbox/pull/226
This PR is backwards compatible with current `ListItem` syntax and enables following customizations:

image

```
<List>
  <ListItem leftIcon='send' rightIcon='done' caption='Reference item'/>
  <ListItem rightIcon='done' caption='Item with custom left icons'>
    <FontIcon value='send' />
    <Avatar image='https://pbs.twimg.com/profile_images/459485216499720192/ufS4YGOY_400x400.png'/>
  </ListItem>
  <ListItem leftIcon='send'>
    <ListItemContent caption='custom right icons' legend='ListItemContent acts as a divider'/>
    <FontIcon value='done' />
    <FontIcon value='undo' />
  </ListItem>
  <ListItem leftIcon='mail' rightIcon='create'>
    <ListItemContent>
      <ListItemText primary> Custom Caption </ListItemText>
    </ListItemContent>
  </ListItem>
  <ListItem leftIcon='save' rightIcon='delete'>
    <ListItemContent>
      <ListItemText primary onClick={() => {console.log('clicked caption');}}>
        Custom caption with events
      </ListItemText>
      <ListItemText> Custom legend with correct height </ListItemText>
    </ListItemContent>
  </ListItem>
  <ListItem caption='Item with a button'>
    <Button icon='save' label='save'/>
  </ListItem>
</List>
```

`FontIcon` and `Avatar` custom components are applied with default styles to match current list style. Everything else, like the `Button` in the example above would have to be styles manually by end user or added to default styles.

@javivelasco  @nathanmarks, what do you think of this implementation?


nathanmarks:
Looks a lot cleaner than the last one! Nice

Quick thing that we need to make sure is possible (it's a relatively common pattern in google apps)

The rows should be easily turned into clickable rows with a ripple, with the ability to link an avatar or an icon independently. I haven't checked this out yet but @kagux can you confirm whether this is possible or not in your implem?

kagux:
@nathanmarks thanks!
I tried following example on a List with `ripple selectable`

```
<ListItem caption='Item with overlayed click events' onClick={() => console.log('clicked row')}>
  <FontIcon value='send' onClick={e => {e.stopPropagation(); console.log('clicked avatar');}}/>
  <Avatar
    image='https://pbs.twimg.com/profile_images/459485216499720192/ufS4YGOY_400x400.png'
    onClick={e => {e.stopPropagation(); console.log('clicked avatar');}}
  />
</ListItem>
```

While you can link avatar/icon seprately from row itself, `Ripple` component overlays the whole row so you see ripples no matter where you click.
I'll see if there is a way to make it work.

nathanmarks:
@kagux Okay cool.

What about z-index that is conditionally applied? Remember, we don't want icons that aren't supplementary actions to not trigger the main list item click handler.

Here's a good example of where the combined functionality is needed:
image

kagux:
@nathanmarks your example makes perfect sense, I agree that it should be possible.

`Ripple` is positioned absolutely and also is a sibling of a parent. That makes z-index irrelevant because of stacking context.

nathanmarks:
@kagux we can wrap the other list contents separately.

ListItem
 - contents
 - ripple

I'm also wondering if a `<button>` should be injected as an additional wrapper in the list item if an `onClick` handler is provided -- this would be semantically correct.

kagux:
@nathanmarks could you elaborate your idea `we can wrap the other list contents separately`?

nathanmarks:
@kagux pseudo-html

```
<ListItem>
  <ListItemContents>
  <Ripple>
</ListItem>
```

nathanmarks:
That's how the current list is too IIRC

nathanmarks:
image

nathanmarks:
@kagux It looks like the actions are already clickable -- so all it would take is preventing the click event bubbling back and triggering the main ripple.

I wonder if we should actually leave this up to the user. What do you think?

kagux:
@nathanmarks but that's what causing the issue right now because `Ripple` is sibling of a parent for every element inside a list. Unless I'm missing something.

DOM from this PR is similar 

image

nathanmarks:
@kagux When you click on the current list, on an icon, it is the the node that triggers the event:

image

nathanmarks:
So yeah, since the list items are composable now, we could leave the event bubbling prevention up to the user. Sometimes it's hard to find the balance between keeping things simple and making it easy to implement common spec examples

kagux:
@nathanmarks you're right, the icon span node triggers event. If I add `onMouseDown={e => e.stopPropagation()}` it doesn't bubble up and no ripple.
We can stop bubbling here https://github.com/react-toolbox/react-toolbox/pull/282/files#diff-b5ed0f8b3f5d36a8a158464c29236615R12 when wrapping components. If we let user mark clickable components with a property like `clickable`/`listItemClickable`, then it would be trivial. That would solve most common use case and users would still be able to take control of event propagation if they wanted to.

nathanmarks:
@kagux perhaps something like `secondaryAction` -- is more expressive than `clickable` and makes it clear the the action is secondary to a primary action and therefore should not bubble the event.

On the other hand, maybe we could leave the prop out for now to keep the API surface small. If someone is wiring up individual icons on lists they are already getting deep enough into things that preventing the bubbling is easy.

I also want to avoid doing **too** much in this specific area as after you've done your PR I want to look into creating some automatic wrappers to make components more semantic output-wise.

For example, having a `<li>` element with an event handler on it would be better served with a `<button>` injected as the immediate child with an aria-label on it.

kagux:
@nathanmarks I've updated PR with small change to stop Click/MouseDown propagation if item action component defines onClick. That would be the most common scenario to add onClick events to action component root. It prevents ripple and row click. If a user wants a more complicated component, then it's up to them to stop propagation.

nathanmarks:
nice, sounds good

javivelasco:
Amazing work guys. Good to go!
Thanks so much both of you!

PR #293, opened 2016-02-04
Update TS Definition

hsrobflavorus:
* Fixed for TypeScript 1.8.0's stricter var declaration requirements (move `declare var ...` inside each individual module).
* Removed triple-slash reference to React to fix npm install compatibility (you'll need to make sure you're referencing react.d.ts somewhere in your project!).
* Fixed default exports where applicable


javivelasco:
Thanks!

PR #298, opened 2016-02-06
Use consistent data attributes

fredguest:
data-toolbox should be data-react-toolbox to be consistent with the rest of the components.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,10 @@ const CustomButton = () => (
 );
 ```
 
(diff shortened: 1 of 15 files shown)


javivelasco:
Agree with you. Maybe in this PR we can add other data attributes that are missing for some components before merging. What do you think?

fredguest:
ya good call, i'll add to this PR.

fredguest:
should be good to go unless you see something missing, or have a reason to prefer data-role over data-react-toolbox.

javivelasco:
Nice! I see it ok :) thanks!

PR #301, opened 2016-02-07
allow app-bar height override

anthu:
As mentioned in [Configuring react-toolbox sass variables with toolbox-loader](http://stackoverflow.com/questions/35255862/configuring-react-toolbox-sass-variables-with-toolbox-loader#comment58226733_35256084) there is no possibility to override the height of the app-bar component.

I added the sass `!default` flag and adapted the padding of the app.


javivelasco:
Thanks! :)

Issue #303, opened 2016-02-07
`findDOMNode` called on unmounted component


emirotin:
The menu component tracks document clicks (I didn't dig deeper, but I think to hide itself). It looks like unmounted component someteimes is still watching for these events, and then the handler calls `findDOMNode` which leads to the exception being thrown.

I have multiple views (chat rooms) each having its own icon menu. 
To reproduce the error:
- open some room
- expand the menu
- click a link to a different view

After that the error is thrown on every link click.

PR #310, opened 2016-02-10
Cleanup listeners on document before unmount

Kl0tl:
If a `Menu` or a `Dropdown` is removed from the DOM while open React throws `Invariant Violation: findDOMNode was called on an unmounted component.` when the next click on the document is intercepted because the listeners are not removed.

I don’t think `Slider`s nor `ClockHand`s throw but there’s still a memory leak as the listeners hold a reference to the component.

### DIFF
--- a/components/dropdown/Dropdown.jsx
+++ b/components/dropdown/Dropdown.jsx
@@ -43,6 +43,12 @@ class Dropdown extends React.Component {
     }
   }
 
+  componentWillUnmount () {
+    if (this.state.active) {
+      events.removeEventsFromDocument({click: this.handleDocumentClick});
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


javivelasco:
Yeah, it makes sense! Please fix linter errors and we are ready to merge. I think this solves #303

Kl0tl:
There’s no linter errors and `npm test` runs fine on my machine but fails on Travis with `Cannot find module 'postcss-normalize-url'`. Could this be a cache issue ?

Kl0tl:
I edited my commit because I forgot about removing the `Slider` keyboard events. Don’t know why but Travis seems happy now !

javivelasco:
Maybe it was circumstantial :D 
Merged!

PR #311, opened 2016-02-11
Move Overlay directly to virtual DOM and fix Dialog overflow

tuckerconnelly:
We're using react-toolbox with redux, and needed the context a Dialog component. It wasn't available because the Overlay was directly to the DOM, outside the virtual DOM. Moved it back into a regular render() method so we could use redux with Dialog components.

Also, within a dialog component, the Autocomplete component was getting cut off because overflow-y was set to auto. Set overflow-y to visible :)

Thanks!

### DIFF
--- a/components/dialog/style.scss
+++ b/components/dialog/style.scss
@@ -43,7 +43,7 @@
 .body {
   flex-grow: 2;
   padding: $dialog-content-padding;
-  overflow-y: auto;
+  overflow-y: visible;
   color: $color-text-secondary;
 }
 
--- a/components/overlay/Overlay.jsx
+++ b/components/overlay/Overlay.jsx
@@ -16,39 +16,18 @@ class Overlay extends React.Component {
     invisible: false
   };
 
-  componentDidMount () {
-    this.app = document.querySelector('[data-react-toolbox="app"]') || document.body;
-    this.node = document.createElement('div');
-    this.node.setAttribute('data-react-toolbox', 'overlay');
-    this.app.appendChild(this.node);
-    this.handleRender();
-  }
-
-  componentDidUpdate () {
-    this.handleRender();
-  }
-
-  componentWillUnmount () {
-    ReactDOM.unmountComponentAtNode(this.node);
-    this.app.removeChild(this.node);
-  }
-
-  handleRender () {
+  render () {
     const className = ClassNames(style.root, {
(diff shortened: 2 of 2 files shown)


kagux:
There was a reason behind having overlay outside of virtual DOM.

 First,  if some parent nodes of component have transformation, then fixed css changes viewport size to those of parent instead of client viewport.  

Second,  if you solve the viewport issue with dynamic position using state,  you'll hit context stack problem.  Parent components and their siblings  with fixed/relative/etc will always be on top of overlay no matter what z-index it has. 
Try running docs site with your changes to see what I'm talking about. 

While it doesn't look like it at first,  in fact current implementation is very elegant if you consider where tackling all edge cases of straightforward solution would take it. Thumbs up to the authors!

tuckerconnelly:
Hmm, I think not having context within the Dialog is a bigger problem than Dialogs being contained within transformed parents. Packages like react-redux and react-intl (and anything else that relies on context) won't work with Dialog then.

Maybe an property to remove from the virtual DOM would be a good idea?

```
<Dialog removeFromVirtualDOM />
```

javivelasco:
Thanks for your words @kagux and your PR @tuckerconnelly 

Components that take the whole viewport can't be holding as regular components since they will not look good because of the reasons @kagux exponed. I agree with you we need to find a way to share the context though. The solution would be, instead of rendering the overlay in a DOM node out of the tree, render inside the tree but as a top element. There is an open issue about this since as you said the context is not shared and, for example, a Drawer can't hold a navigation menu relying on react router.

patrick-jones:
Thinking out loud a bit here... I think @javivelasco is correct that we place responsibility for the overlay in a component that is typically shown at full screen. What if this component then places a service object on the react context with a `showWithOverlay(Component, [options])` function (along with `close(result)`).  Components such as `Dialog` could access the API through the context and ask the higher component to render their contents within the overlay.

In fact, if `showWithOverlay` returned a `future`, you could have a nice flow:

```javascript
overlayService.show(MyComponent).then(result => hooray(result)).catch(reason => tooBad(reason));
```

javivelasco:
@patrick-jones sounds nice, we can test an approach like that!

patrick-jones:
I made a little [proof of concept](http://codepen.io/patrick-jones/pen/gPNqEM) to expose a dialog service through the context.

PR #312, opened 2016-02-12
type -> getType as `type` seems to be a keyword in node 5.1

jackielii:
without this, it shows error when starting:
```
ERROR in ./components/list/ListItemContent.jsx
Module build failed: SyntaxError: /home/jackieli/nodejs/react-toolbox/components/list/ListItemContent.jsx: "type" is read-only (This is an error on an internal node. Probably an internal error. Location has been estimated.)
```


javivelasco:
Thanks!

PR #317, opened 2016-02-15
dropdown accepts string or number as value

epilgrim:
I couldn't find any reason to restrict the type of the values passed to dropdown to strings. Using integers is common when dealing with ids


epilgrim:
My first though was to set the validation to `React.PropTypes.any`, same way as it is in Input component. But then I thought that perhaps you wanted to be more conservative, and thus I changed it to either string or number.

Let me know if you are fine with any, and I will update the PR

javivelasco:
Agree, thanks :)

Issue #251, opened 2016-01-07
weird issue around IconMenu


jackielii:
Used in combination with AppBar etc the width and height of the IconMenu is wrong, in my application with width is simply the width of the viewport.

Test code here: https://github.com/react-toolbox/react-toolbox/compare/dev...jackielii:issue-251

![menuiconerror](https://cloud.githubusercontent.com/assets/360983/12175658/602035fc-b55b-11e5-8e47-aeb577d8370a.png)

It should look like this (copied from official doc)
![image](https://cloud.githubusercontent.com/assets/360983/12175803/13a03fd2-b55c-11e5-99f4-31bd84d164c2.png)

javivelasco:
Yes, probably is related to the fixed position of the container element. Maybe the best would be to position the menu on a top node like we do with other components like the `Drawer` or the `Dialog`. Anyway, it's a bug and case to study, thanks for reporting it!

jackielii:
I think we could achieve a Menu with a `Diaglog` and `List`, `LstItem` etc. right?

javivelasco:
Yep

erdysson:
No I think it is not related to the fixed position of the app-bar, because I have the  same issue in both appbar and inside the body

sars:
Position is defined on componentDidMount
it seems like getBoundingClientRect returns wrong info in that time
If you wrap calculations in componentDidMount with setTimeout - it works good. Otherwise - not

quarryman:
Got the same behaviour. SetTimeout with zero delay solves the issue.

PR #323, opened 2016-02-17
race condition between menu item layout and container size assessment

Blechhirn:
Fixes https://github.com/react-toolbox/react-toolbox/issues/251 as suggested by @quarryman

### DIFF
--- a/components/menu/Menu.jsx
+++ b/components/menu/Menu.jsx
@@ -43,9 +43,11 @@ class Menu extends React.Component {
   };
 
   componentDidMount () {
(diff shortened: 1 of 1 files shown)


javivelasco:
Yeah, I think it's the right thing to do in this case

sars:
this solution looks a little bit hacky, but i couln't find anything better however...
just interesting, why getBoundingClientRect in componentDidMount returns bad value?

javivelasco:
I'm curious about it too. I did not dig too much on it, but possibly it's because the container is animated and it's being moved? I don't know :S

Issue #296, opened 2016-02-05
Add the ability to change the table head titles


derekstavis:
Currently, `<TableHead>` uses model keys as the title of table headings. Adding the ability to change the title in the model would be nice. My proposal is that a new key `title` is added to the model, and is used when available. When not available, fallback to using object key. I can work on such PR, if you do find this interesting upstream.

javivelasco:
I think its a good enhancement with no breaking changes. I'd be happy to accept a PR about it!

Issue #268, opened 2016-01-20
Ability to add buttons to a Table


delaneyj:
I have a table of games which I want players to be able to join or create a new one.  Normally I'd add an extra `<th><Button>create</Button></th>` and in each row add a `<td><Button>join</Button></td>` at the end.  For a lot of table like data this workflow makes a lot of sense but with the current data driven approach in `<Table/>` it appears its not possible (or not documented).

Was hoping something way to include arbitrary components in `td` or `th` elements.

As a stop gap I'm using a normal table and extend the sass in `~react-toolbox/lib/table/style.scss`.  but lose out on the other features of the included `<Table/>`

delaneyj:
As a stop gap I'm using a normal table and extend the sass in `~react-toolbox/lib/table/style.scss`.  but lose out on the other features of the included `<Table/>`

javivelasco:
Yes, it's happening with other components as well. The ability to provide arbitrary content is very important and I think it is specially for components such as the `Table` Although sometimes is very handy to be able to generate a table from a data set, you are also losing flexibility if you want to provide arbitrary content. It's possible anyway.

I'm going to label as an enhancement. I appreciate help from the community with this features, because they actually appear with usage and I'm not enjoying too much spare time lately.

delaneyj:
I actually think Dropdown is the perfect example for this, were the default data is handled fine but can provide a custom `th` & `td` templates.  Would be happy to provide some jist as I use it more, right now using the styling is enough for my use case but we can discuss further in a few weeks.

javivelasco:
Great, thanks!

joehua87:
Thank you for your awesome library.
@javivelasco, @delaneyj I've implemented this feature at https://github.com/joehua87/react-toolbox/commit/86ce8cce2547fdbb00674264963ff959cdc5d308. Please take a look at :smile: I'm very happy to make a PR
Live demo here: http://45.79.94.169:8080

delaneyj:
looking good, can you version also support buttons in the <th/>'s of the header @joehua87 ?

kagux:
Nice work @joehua87 ! 
However, I feel like proposed implementation is introducing quite complicated templating instead of using React components. For now you wanted to customize single cell, but imagine adding custom logic to header or specific rows (colspans, highlighting, etc).
Wouldn't it be better to allow using components that follow original html table markup? Here's an advanced example from [material-ui](http://www.material-ui.com/#/components/table)
```
<Table
[...]

PR #324, opened 2016-02-17
[Table] enhancements

Kosta-Github:
- allow usage of `react` components in table cells (header and rows): https://github.com/react-toolbox/react-toolbox/issues/268
- allow changing table header titles (via `title` prop in `model`): https://github.com/react-toolbox/react-toolbox/issues/296
- allow specifying `onChange` handler per column (via `model`) in order to make only certain columns editable instead of the whole table

Sample usage can be seen in the update `Table` test file.

### DIFF
--- a/components/table/Table.jsx
+++ b/components/table/Table.jsx
@@ -69,7 +69,7 @@ class Table extends React.Component {
           index={index}
           key={index}
           model={this.props.model}
-          onChange={this.handleRowChange.bind(this, index)}
+          onChange={this.props.onChange ? this.handleRowChange.bind(this) : undefined}
           onSelect={this.handleRowSelect.bind(this, index)}
           selectable={this.props.selectable}
           selected={this.props.selected.indexOf(index) !== -1}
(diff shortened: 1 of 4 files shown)


javivelasco:
Let me check it this afternoon :) Great work

epilgrim:
Nice one! Allowing to pass ReactComponents for fields and headers makes total sense.
As we go that way, wouldn't' make sense also deprecate the magic around auto-editable fields?

now, we can pass custom components. Why not let those components to be bound to onChange events, or onclick, or whatever. The table is supposed to display in a tabl-y way. All the other behaviour should be on each passed component. We can pass now any Input, Autocomplete, select, etc.

Anyways, if you don't want to deprecate that behavior (for BC), I don't think that adding the onChange by column is necessary anymore, for the same points stated above.

What do you think?

Kosta-Github:
I agree with @epilgrim that the cell content should be responsible for doing the proper `onChange` handling.

But I wanted to make as little changes as possible to achieve all 3 goals listed above without introducing breaking changes to existing code.

javivelasco:
Very useful, thanks!

PR #325, opened 2016-02-18
Add `valueLink` to Input

LestaD:
https://facebook.github.io/react/docs/two-way-binding-helpers.html

### DIFF
--- a/components/input/Input.jsx
+++ b/components/input/Input.jsx
@@ -41,14 +41,14 @@ class Input extends React.Component {
   renderInput () {
     const {multiline, value, valueLink, ...others} = this.props;
     const className = ClassNames(style.input, {[style.filled]: value});
-    let props = {
+    const props = {
       ...others,
       className,
       ref: 'input',
       role: 'input'
     };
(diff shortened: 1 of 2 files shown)


oliviertassinari:
`valueLink` is going to be removed from the react core. I wouldn't go into this direction.

javivelasco:
Agree with @oliviertassinari Two way binding helpers sound like an antipattern in react from my POV.
Do you have a specific scenario where you are using them and they can't be removed? Removing `valueLink` from core means a lot

LestaD:
@oliviertassinari 
> ReactLink is just a thin wrapper and convention around the onChange/setState() pattern. It doesn't fundamentally change how data flows in your React application.

Where you read about removing `valueLink` from react core? proof?

oliviertassinari:
@LestaD  https://github.com/facebook/react/blob/master/src/renderers/dom/client/wrappers/ReactDOMInput.js#L98

LestaD:
@oliviertassinari oh!

PR #326, opened 2016-02-18
Add Positions for dialog actions

kodermax:
Please add this feature.


javivelasco:
Thanks for the PR!

May you check the spec page in the repository and the docs? I think it's possible to set a position for actions without actually adding new props for it, just styling a little bit. Being that the case, do you really think it's better to add new props?

Anybody else has an opinion on this?

javivelasco:
Closing meanwhile. Check the docs, you should be able to do so without changing any code.
Thanks for your PR! 😄

PR #327, opened 2016-02-18
Add prop size for CardTitle

kodermax:


javivelasco:
Sounds reasonable but docs and spec should be updated as well before merging :)

PR #329, opened 2016-02-18
Dropdown doesn't close on click outside fix

sars:
Hi there,

Dropdown component doesn't close on click outside in my app. 
I debugged a little bit problem and found that componentWillUpdate uses incorrect.
It accepts nextProps and nextState and not prevState and nextState
https://facebook.github.io/react/docs/component-specs.html#updating-componentwillupdate

### DIFF
--- a/components/dropdown/Dropdown.jsx
+++ b/components/dropdown/Dropdown.jsx
@@ -31,8 +31,8 @@ class Dropdown extends React.Component {
     up: false
   };
 
-  componentWillUpdate (prevState, nextState) {
-    if (!prevState.active && nextState.active) {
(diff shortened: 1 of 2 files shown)


epilgrim:
Correct, you are totally right!!

Can you also update the File components/menu/Menu.jsx, as that's the one I used as base, and incurres in the same error?

javivelasco:
Probably we changed the method name and forgot to update the arguments. Can you do as @epilgrim says and update the `Menu` too before merging? :dancer:

sars:
@javivelasco , done!

javivelasco:
Great, thanks! :)

PR #330, opened 2016-02-18
renderInput() method never used

sars:
renderInput() method never used
or i'm mistaken?


javivelasco:
You are right, probably I missed it during a refactor, thanks!

Issue #328, opened 2016-02-18
TimePicker view problem


sars:
Hi there,

Thanks guys for this library.

I have one problem with TimePicker that I don't know hot to fix
On the first dialog open, clock looks like 

<img width="226" alt="screen shot 2016-02-18 at 5 30 50 pm" src="https://cloud.githubusercontent.com/assets/226194/13148480/3b4899a0-d666-11e5-9558-35d1810c1ae3.png">

and then immediately changes to 

<img width="316" alt="screen shot 2016-02-18 at 5 29 54 pm" src="https://cloud.githubusercontent.com/assets/226194/13148481/3b4a41ec-d666-11e5-910b-38e80536f368.png">

Could you suggest me how to fix this?

javivelasco:
Wow. Can you provide more information about the environment? Browser and version, if you customized the dialog with a new animation or something similar to reproduce it?

BTW, thanks for your kind words :)

sars:
Chrome, last version,
I didn't customise dialog or picker component...
I didn't find the reason so far...

sars:
hmmmm... may be this is related to SSR...
when i go from another page to target page with timePicker - if works fine
when i go directly to the target page (SSR rendering happens) - it works bad

PR #332, opened 2016-02-19
Clock SSR fix

sars:
this fixes https://github.com/react-toolbox/react-toolbox/issues/328
I'm not sure it is the best solution, but it works

### DIFF
--- a/components/time_picker/Clock.jsx
+++ b/components/time_picker/Clock.jsx
@@ -30,7 +30,9 @@ class Clock extends React.Component {
 
   componentDidMount () {
     window.addEventListener('resize', this.handleCalculateShape);
-    this.handleCalculateShape();
+    setTimeout(() => {
+      this.handleCalculateShape();
+    });
   }
 
   componentWillUnmount () {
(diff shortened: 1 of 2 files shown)


sars:
this is really similar to recently merged https://github.com/react-toolbox/react-toolbox/pull/323/files

components/time_picker/Clock.jsx
```diff
@@ -30,7 +30,9 @@ class Clock extends React.Component {
 
   componentDidMount () {
     window.addEventListener('resize', this.handleCalculateShape);
-    this.handleCalculateShape();
+    setTimeout(() => {
+      this.handleCalculateShape();
+    });
```
Kosta-Github: I am not sure if this could ever happen, but what if this component get unmounted again before the `timeout` trigger is fired?

you might need to do something like this:
```JSX
componentDidMount() {
    ...
    this._cancelTimeout = setTimeout(() => this.handleCalculateShape());
}
componentWillUnmount() {
    ...
    cancelTimeout(this._cancelTimeout);
    this._cancelTimeout = undefined;
}
```
@Blechhirn

javivelasco:
Apparently it's a common issue when a component needs to check the DOM to calculate sizing... It works so let's merge. I'm going to release a new version with all these fixes during the weekend

sars:
@javivelasco thanks

PR #333, opened 2016-02-19
Created DynamicAutocomplete field.

dhamilton91:
Hey guys,

This is a new Dynamic Autocomplete component. 
It is used to dynamically populate the suggestions so that a service can be called to return results.

David

### DIFF
--- a/components/dynamic_autocomplete/DynamicAutocomplete.jsx
+++ b/components/dynamic_autocomplete/DynamicAutocomplete.jsx
@@ -94,7 +94,11 @@ class DynamicAutocomplete extends React.Component {
                 </li>
             );
         });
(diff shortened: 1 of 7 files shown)


javivelasco:
Hi! What do you mean by Dynamic? You can always change the source prop and the autocomplete will change as well

javivelasco:
Also, this is pretty much duplicating code :(

dhamilton91:
You can update the source dynamically but you cant easily use the value from the input to use as a service call parameter. Also the current autocomplete field does some extra filtering which is unnecessary in this situation.

kagux:
I think a dry approach would be allowing to pass optional search function as prop.

PR #339, opened 2016-02-22
Added a few more events to Dialog's overlay

necinc:
I make an image croper using dialog component, and I need to registrate onMouseUp event on the all visible part of the document, so overlay takes it.  That's why i decided to add a few event to overlay which allow me to complete it. It may be helpful for someone.


javivelasco:
It makes sense! Can you also add to the documentation the new props?

necinc:
Done.

javivelasco:
Thanks!

PR #344, opened 2016-02-23
Change error PropType on Input to node

tuckerconnelly:
We had a weird use case where errors were coming back from the API as HTML.

So to display the errors we did this:

```js
<Input error={
    <span dangerouslySetInnerHTML={{ __html: errors[name] }} />
  }
/>
```

But the error PropType is string, so it was throwing an error. Changed it over to node :)


javivelasco:
Is this still valid when you use a string? If not, I think both should be valid!

tuckerconnelly:
Yep, both work

javivelasco:
Alright

PR #351, opened 2016-03-01
Parameters onChange and onSelect table line brought to the top level

serslon:


javivelasco:
Thanks for your PR!

What is this exactly solving? It is adding 23 lines but I can't see exactly what is providing

serslon:
Sorry but I closed this pull request because work not ended and I don't have more time.
This change accept disabled editable fields or much more actions for field. But styles broken and need fixed style also.

PR #352, opened 2016-03-03
When the font-size (or $unit) is small, there's no difference between selected and unselected item

zalmoxisus:


javivelasco:
May you add a screenshot or something to illustrate the problem and solution?
Thanks for the PR!

zalmoxisus:
Surely, as you see in [the first screenshot](https://github.com/zalmoxisus/social-expert#social-expert) the selected item doesn't have a different font weight due to small font size, so we differentiate it by adding a background color.

javivelasco:
Alright! I'd be very happy if you mention React Toolbox as well in the Readme!
It's a very nice project BTW, congrats!

zalmoxisus:
@javivelasco, added. Thank you for the great tools!

javivelasco:
Thank you for your mention and PR @zalmoxisus ! :)

waxmihs2902:
good thing hapen to good people like you.. lets pag it forward! cheers! thanks tol all!

PR #358, opened 2016-03-06
Fix label for number inputs with 0 as value

TobiasBales:
For an
`<Input type="number" label="label text" value={0}/>`
the label would overlay the value when it was set to 0 (the integer)
since the checking logic checked for javascript "trueishness".
Now the check is explicit against null, undefined, empty string and NaN.


TobiasBales:
@javivelasco Any reaction on this?

javivelasco:
Missed it. Totally makes sense, thanks!

PR #363, opened 2016-03-08
Fixes small typos in AutoComplete documentation

suancarloj:


components/autocomplete/readme.md
```diff
@@ -29,7 +29,7 @@ class AutocompleteTest extends React.Component {
         label="Choose countries"
         onChange={this.handleChange}
         source={source}
-        value={this.state.value}
+        value={this.state.countries}
```
javivelasco: Yes, there is a mistake here. Also we should change the change handler. It is setting a value state instead of countries

suancarloj: Fixed! :)

javivelasco: Thanks!

PR #364, opened 2016-03-08
Modified z-index to include default. 

### LINKED ISSUE #362: Adding !default to z-index values in _global.scss
Is it possible to add the !default flag to the z-index values in _global.scss? I need to overwrite them and according to a stack overflow post, it said that the _global.scss variables should have the !default attribute.

KerenChandran:
As mentioned in #362, this PR contains the !default flag for z-index values.

Fixes #362.


javivelasco:
Thanks! :D

Issue #320, opened 2016-02-17
DatePicker other language


kodermax:
How can i change language for DatePicker?

javivelasco:
Not for now. But it can be a good improvement to provide a custom array months as a prop

felipeleusin:
Would you guys be interested in using moment or Intl support? If not, I think just providing a format method that defines how to display the date would already help.

javivelasco:
As few dependencies needed the best IMO. Also I think the format to display the date in a picker is specified in Material Design isn´t it?

felipeleusin:
I couldn't find anything in the spec regarding the format. Imho, it should be localizable. Regarding the dependency Intl has less dependencies but needs polyfill for some browserd

PR #366, opened 2016-03-08
Allows to format the date displayed on datepickers input

felipeleusin:
This allows to display a custom format after selecting a date on the DatePicker. It defaults for the current solution.

It's a first step in order to be able to localize the DatePicker, while I agree that taking a dependency on something such as Intl or moment maybe too much, we can provide some better options such as allowing to format the Dialog display date.

### DIFF
--- a/components/utils/time.js
+++ b/components/utils/time.js
@@ -179,7 +179,7 @@ const time = {
 
   formatDate (date) {
     return `${date.getDate()} ${time.getFullMonth(date)} ${date.getFullYear()}`;
-  },
+  }
 };
 
 export default time;
--- a/components/utils/time.js
+++ b/components/utils/time.js
@@ -1,4 +1,4 @@
-export default {
+const time = {
   getDaysInMonth (d) {
     const resultDate = this.getFirstDayOfMonth(d);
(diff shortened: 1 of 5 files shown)


components/utils/time.js
```diff
@@ -1,4 +1,4 @@
-export default {
+const time = {
```
javivelasco: I think this is just a matter of code style. Any particular reason to move the exposing to an object?

javivelasco:
I like the approach, thanks! May you make Travis pass so we can merge?

javivelasco:
Also add the `fixes` in the comments to auto-close the related issue

components/utils/time.js
felipeleusin: I actually had to do this because on https://github.com/react-toolbox/react-toolbox/pull/366/files#diff-b92a8081cc5e0d62646a9cd36e0f58dcR181 I had to reference time

felipeleusin:
Do you mean #320 ? I don't think this fixes it, since what they are asking if for full localization. I can take this up but first I want to take a shot at improving the autocomplete.

Updated the PR and travis is passing.

javivelasco:
You're right. Thanks for the PR! :)

PR #370, opened 2016-03-09
fixes #369

### LINKED ISSUE #369: Autocomplete: disabled property does not disable selected items
For [example:](http://react-toolbox.com/#/components/autocomplete)
```
 <Autocomplete
        direction="down"
        disabled
        label="Choose countries"
        name="countries"
        onChange={this.handleChange}
        source={source}
        value={this.state.countries}
      />
```
This will disable the input on the autocomplete, but will not disable the selected items at the top if it. So if I were to click on an item, it would still get deleted even when disabled was set to true. The simple fix is to add a check to see if disabled is set to true before calling the ```unselect``` function on an item.

arhamahmed:
Added a check to see if the disabled prop is set to true before unselecting an item.

### DIFF
--- a/components/autocomplete/Autocomplete.jsx
+++ b/components/autocomplete/Autocomplete.jsx
@@ -142,9 +142,11 @@ class Autocomplete extends React.Component {
  }
 
  unselect (key, event) {
-   const values = this.values(this.props.value);
-   values.delete(key);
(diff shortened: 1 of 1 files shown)


javivelasco:
Great! Thanks :)

Issue #292, opened 2016-02-03
Hint text for inputs?


tnrich:
Hey there,

I didn't see the option to supply a hint text to react-toolbox Input elements: http://react-toolbox.com/#/components/input

Does this option exist, and if not, is there a workaround to get hint texts?

Thanks!

KerenChandran:
You can use the placeholder or label property.

<Input type='text' label='Name' name='name' value={this.state.name} onChange={this.handleChange.bind(this, 'name')}  />

label will be initially placed on the field and then go to the top when the field is selected.
placeholder will act like it does on a normal input field, ie. it will only be visible if the field doesn't contain text.

The problem is you can't use both as they initially overlap, but when the field is active, the label is pushed to the top and behaves like it should.

tnrich:
Hmm thanks for the response. That seems like a bug to me. Shouldn't you be
able to use both a label and a hint text?

KerenChandran:
Ya I was surprised by this too. Material-ui has that feature built in on their demo page, so I believe this is a bug. See 'Floating Label' field: http://www.material-ui.com/#/components/text-field

javivelasco:
As far as I know, the Material Spec does not have a placeholder consideration. I guess we can tweak this a bit to display a placeholder when the component is focused. As @KerenChandran said, you can do it now but it overlaps with the label.

tnrich:
I think that would be a nice improvement @javivelasco  +1 :)

arlair:
I just found this as well and was going to report it.

Cool project and finding the code easy to read. Nice job :).

javivelasco:
Thanks @arlair :) Gonna label it as enhancement.
As I always say, help here is appreciated!

PR #371, opened 2016-03-10
Added hint attribute to input field.

KerenChandran:
References #292 

Since no one was working on it, I thought I could give myself a challenge and add the hint text. I'm new to SCSS, so if there's a mistake then it would be great if you can tell me what I did wrong.

I also updated readme.md for input and added two new fields on the test page to show off the hint attribute in action (one with custom icon and another with multiline).

### DIFF
--- a/components/input/Input.jsx
+++ b/components/input/Input.jsx
@@ -79,7 +79,7 @@ class Input extends React.Component {
         {icon ? <FontIcon className={style.icon} value={icon} /> : null}
         <span className={style.bar}></span>
         {labelText ? <label className={labelClassName}>{labelText}</label> : null}
-        {hint ? <label className={style.hint}>{hint}</label> : null}
+        {hint ? <span className={style.hint}>{hint}</span> : null}
         {error ? <span className={style.error}>{error}</span> : null}
         {maxLength ? <span className={style.counter}>{length}/{maxLength}</span> : null}
         {children}
--- a/components/input/Input.jsx
+++ b/components/input/Input.jsx
@@ -10,6 +10,7 @@ class Input extends React.Component {
     disabled: React.PropTypes.bool,
     error: React.PropTypes.node,
     floating: React.PropTypes.bool,
+    hint: React.PropTypes.string,
     icon: React.PropTypes.any,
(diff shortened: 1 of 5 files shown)


components/input/Input.jsx
```diff
@@ -77,6 +79,7 @@ class Input extends React.Component {
         {icon ? <FontIcon className={style.icon} value={icon} /> : null}
         <span className={style.bar}></span>
         {labelText ? <label className={labelClassName}>{labelText}</label> : null}
+        {hint ? <label className={style.hint}>{hint}</label> : null}
```
javivelasco: This is the only thing that sounds a little weird to me. Having a `label` tag element wrapping the hint text does not sound very semantic IMO. Shouldn't it be a span or something similar?

KerenChandran:
You are right. I used the label as a starting point cause it had a similar behaviour to the hint attribute. I'll submit the change to a span in a few hours.

javivelasco:
Alright!

KerenChandran:
Hm... I don't understand as to why a simple change from label to span will cause npm install to fail. I even ran npm run test in my terminal and it passed all tests. Can you please help me look into this issue with Travis?

KerenChandran:
Hm.. After looking at #376 and checking the docs, it appears the hint/helper text may need to appear below the input field. Should I modify the code to behave like that or leave it as is?

https://www.google.com/design/spec/components/text-fields.html#text-fields-required-fields

renato:
The Text Field spec describes a "helper text" below the field one single time, in the required field section. All other sections in the same page puts a "hint text" inside the text field.

I think it's right to expect it to be inside the field.

javivelasco:
Since it's an easy change, I think it causes to pain to keep it as it is right now. Sometimes MD looks confusing for me 😐

Thank you both!

KerenChandran:
Fixes #292 

Sorry should have mentioned this before when submitting the ticket. Hopefully this closes the issue, otherwise @javivelasco can you close the issue?

javivelasco:
Sure, thanks @KerenChandran

PR #372, opened 2016-03-10
Provide own class name to Input of DatePicker

necinc:
We often want to style the `DatePicker`'s `Input` as other inputs component, but we don't have the possibility to pass its some class name, now we can :)

Firstly I've created a new field in props to pass class name to `DatePickerDialog` and pure `className` provided to `Input` element, but it can break the markup in some projects which already use React-Toolbox. That's why I just created the new field to `Input`'s class name without changes in other parts of `DatePicker` component.

### DIFF
--- a/components/date_picker/DatePicker.jsx
+++ b/components/date_picker/DatePicker.jsx
@@ -41,7 +41,7 @@ class DatePicker extends React.Component {
     return (
       <div data-react-toolbox='date-picker'>
         <Input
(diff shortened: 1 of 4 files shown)


necinc:
Sorry but here is too many mistakes, I will fix them soon

javivelasco:
Ok! :)

javivelasco:
Done already

PR #373, opened 2016-03-10
Correctly handle zero as dropdown value

TobiasBales:
This is only an issue if the dropdown item with the value 0
is not the first item still better to properly handle it I think.


TobiasBales:
Somewhat unrelated - any idea on when the next release will happen?

javivelasco:
Thanks! We released a new version with patches the past weekend if I'm not wrong!

Issue #380, opened 2016-03-13
Remove Error Styling on focus for Input Component


KerenChandran:
According to Google's docs for text field, required section, the error styling does not appear when the input field is in focus. Will submit a PR to fix this issue.

PR #381, opened 2016-03-13
Removed Error styling when input field is in focus

KerenChandran:
References #380


javivelasco:
Being that the case it´s alright!
Thanks for the PR!

Issue #383, opened 2016-03-13
TypeError: Cannot read property 'getBoundingClientRect' of undefined


alihammad-gist:
I keep getting this error this is the line
https://github.com/react-toolbox/react-toolbox/blob/dev/components/tabs/Tabs.jsx#L55

```tsx

export default class Sections extends React.Component<props, state> {
	state = {
		tabIndex: 0 
	}

	private onTabChange = (id: number) => {
		this.setState({tabIndex: id})
	}

	render() {
		return (
			<Tabs index={this.state.tabIndex} onChange={this.onTabChange}>
			{this.props.sections.map(section => (
				<Tab 
					label={section.label} 
					key={section.id}
					onActive={() => section.onActive(section.playlistsId)}>
					{section.playlists}	
				</Tab>
			))}
		 	</Tabs>	
		);
	}
}
```

`Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined(anonymous function) @ Tabs.js?9eb4:89`

PR #385, opened 2016-03-16
Fix #361 cancel timer when tabs component unmounts

### LINKED ISSUE #361: Tabs: "Cannot read property 'getBoundingClientRect' of undefined"
Seems to happen when the component starts unmounting (or when it's not fully mounted yet?)
Repro: have 2 pages with tabs, cross-linked. Switch between them.

sosaucily:
If the component quickly mounts and unmounts, the timers will continue running in the background on a component that isn't in the dom.

Need to clearTimeout before calling another setTimeout to ensure there's only 1 running at any time.

https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html

This fixed #361 #383

### DIFF
--- a/components/tabs/Tabs.jsx
+++ b/components/tabs/Tabs.jsx
@@ -27,6 +27,10 @@ class Tabs extends React.Component {
     this.updatePointer(nextProps.index);
   }
 
+  componentWillUnmount () {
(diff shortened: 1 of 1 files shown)


javivelasco:
Great! Thanks!

PR #388, opened 2016-03-17
feat(ListItem): add className prop support

mgcrea:
Needed for custom styling of the ListItem (notably if you want to use flex styles to combine it with a MenuItem floating on the right).

Not sure if I have to change something in the specs (the docs seemed to be already OK). We'll gladly update the PR if I have to.


javivelasco:
No need to change anything else!
Thanks :)

PR #391, opened 2016-03-20
IconButton onMouseLeave now calls the right function

### LINKED ISSUE #389: Tooltip on IconButton doesn't hide
The Tooltip doesn't hide when leaving the mouse from an IconButton.

Tested exactly the same code using a Button and it works.

arhamahmed:
fixes #389

### DIFF
--- a/components/button/IconButton.jsx
+++ b/components/button/IconButton.jsx
@@ -45,7 +45,7 @@ class IconButton extends React.Component {
       className: classes,
       disabled: this.props.disabled,
       onMouseUp: this.handleMouseUp,
-      onMouseLeave: this.handleMouseUp,
+      onMouseLeave: this.handleMouseLeave,
       'data-react-toolbox': 'button'
     };
 


javivelasco:
Thanks!!

tungv:
+1 @arhamahmed 
@javivelasco please get this to npm as I'm blocked by this

components/button/IconButton.jsx
```diff
@@ -45,7 +45,7 @@ class IconButton extends React.Component {
       className: classes,
       disabled: this.props.disabled,
       onMouseUp: this.handleMouseUp,
-      onMouseLeave: this.handleMouseUp,
+      onMouseLeave: this.handleMouseLeave,
```
tungv: but `this.handleMouseLeave` is undefined

tungv: should be `onMouseLeave: this.props.onMouseLeave,` @javivelasco

arhamahmed:
Ah, I overlooked that. #413 should fix this.

PR #393, opened 2016-03-20
Autocomplete single object selection

### LINKED ISSUE #354: AutoComplete Single Select Field State with Object Source
Related to #252 

TLDR: Autocomplete for multiple={false} doesn't read from keys for selecting the entry on the list. Would be nice to fix it to only read from keys and always display the value associated to the key. The code I pasted below is a workaround to use the autocomplete with multiple={false}

Basically from the example below, I modified it to be a single select by adding the attribute multiple={false}. The initial state must be a key as the source code appears to be reading from the list of keys. 

When selecting a new value, the key associated to the value returns (which is good and should be expected). However, re-rendering the component causes the key to be displayed, as it seems that the code ignores looking up the value. This is a problem and so I thought it was a design implementation to go by values for single select autocomplete.

But wait! With a single select, setting the initial state to be the value associated to the key causes another error:
```javascript
TypeError: Cannot read property 'toLowerCase' of undefined
```

However, after the initial state being the key, setting the state of the component to be the value makes it work.

Please fix this or give me other suggestions. 

```javascript

const source = {'ES-es': 'Spain', 'TH-th': 'Thailand', 'EN-gb': 'England', 'EN-en': 'USA'};

class AutocompleteTest extends React.Component {
  state = {
    countries: 'ES-es'
  }

  handleChange = (value) => {
    this.setState({countries: source[value] });
  };

  render () {
  	console.log('state', this.state);
    return (
      <Autocomplete
        direction="down"
        label="Choose countries"
        name="countries"
        multiple= { false }
        onChange={this.handleChange}
        source={source}
        value={this.state.countries}
      />
    );
  }
}

return <AutocompleteTest />

```

KerenChandran:
Fixes #354 

Added a check on cwrp handler to retrieve the query value if the multiple is false. This made the query state to be set as the value instead of the key so the value will be rendered and the key can be stored.

I also fixed the spacing in that file. It was originally indented with a single space and I noticed that the input.jsx file I worked on a few commits ago was indented with two spaces, so I made this match that.


javivelasco:
Is there anything changed with the spacing?? Looks like too many changed lines :/

KerenChandran:
The first commit showed my fix, the second commit was just a spacing fix with no other changes. Will push fix the conflict

KerenChandran:
If you want, I close this and open a new issue with only the fix and no space change

javivelasco:
Ok!

grammka:
@KerenChandran Did u forget to add `query: this.query(nextProps.value)` in `componentWillReceiveProps` to dev branch or I miss smth ?

KerenChandran:
@grammka Hey sorry for not replying earlier. 9d2fe8f has the change you requested.

PR #394, opened 2016-03-20
Allows to close dialog (and all overlays) when pressing esc

### LINKED ISSUE #365: Dismiss date picker with ESC key
Could it be possible to dismiss the date picker when pressing the esc key ?

felipeleusin:
This includes a handler on the dialog and overlay components that captures the Esc key down when the dialog is open. Had to resort to the regular dom addEventListener because I was failing to capture the ESC key without access to body.

Closes #365


javivelasco:
Great!! Thanks @felipeleusin

PR #395, opened 2016-03-20
fixes datepicker autoOk proptype

felipeleusin:
Typo on the propTypes for the autoOk property


javivelasco:
Already fixed, thanks!

Issue #252, opened 2016-01-08
Autocomplete component with multiple=false


dariuszpaluch:
When I use autocomplete with multiple=false (User can choose only one option), I found some mistakes.
1. When we have some selected option, Then user want to change this. Now he need to remove all text from input, then choose onother option from list. I think the best option is: 
   -Remove text from input, and show all options when user click on input (like a simple select)
2. When I select some option from autocomplete, then i remove all from input to choose new option. Then I can't choose this option, which I select before, because it doesn't show in list.

javivelasco:
Yes, I agree with you. It's not a bug but it could work better avoiding the points you've said. More feedback about a desirable behavior would be very nice.

dariuszpaluch:
I think the best option, is add one props to component which describe that clean input on click or not,
and show all options in list when multiple = false.

PR #399, opened 2016-03-23
Renders Selected Item when multiple=false

KerenChandran:
Related #252 

This commit will render the selected value only when autocomplete has multiple=false. I noticed this was missing and I think its a good to have feature. Also when the filters the list to a single item, the suggestions will still be rendered so they can click on that, making it a better user experience.

If there's anything you would like me to consider or fix, please let me know and I'll work on it ASAP.


javivelasco:
I think it makes more sense this way. Let's merge, easy change.
Thanks!

PR #400, opened 2016-03-23
Introduce close() method

kasperekt:
This commit adds `close()` method to Dropdown component. This helps in situation where you must close other dropdowns when one is open.

I've added example in `spec` file.


javivelasco:
Makes sense, thanks!

PR #408, opened 2016-03-25
Fixed checking if key exists

grammka:
I got problem with updating view value in Autocomplete input when changing it with selecting from drop list - the value updating only from second select or triggering setState twice.. I do not quite understand why It was, but these changes fixed it.. btw I think its better to check this way)


KerenChandran:
Hm... Can you please describe your issue? Are you using multiple=false and passing it an object source? If so, I submitted a PR for that issue (#406). When multiple is set to false, the component will receive a new key and will display it directly instead of retrieving the label. I fixed it by calling the query function on nextProps.value in cwrp.

grammka:
Yes, I'm using multiple=false. If default value is undefined and u try to select item, then in this code:
`
query (key) {
   return !this.props.multiple && this.props.value ? this.source().get(key) : '';
 }
`
`this.props.value` is **undefined** and it returns **''**..

KerenChandran:
Oh ok now I see the issue! Its as you said, with this.props.value initially undefined, the code will return '' and still set the value to whatever the key was (as the query function on line 59 is called before the onChange handler passed into the component). Then on the second render, it will correctly set the value you chose since this.props.value is no longer null.

Totally different from #406. This is a good fix as key contains the updated value from the component, whereas this.props.value has not been updated when the call is being made.

javivelasco:
This went soo crazy! I reviewed the code and fixed it. Thanks!

Issue #124, opened 2015-11-17
Table example is not working


TheDeveloperXYZ:
The table example is not working in all browsers (IE, FF, Chrome - Win 8.1).

Throws error in console.

http://react-toolbox.com/#/components/table

javivelasco:
Wow, we forgot to upload the new example. Thanks for reporting, fixing!

ipsips:
it's still broken

`TypeError: unknown: Property id of VariableDeclarator expected node to be of a type ["LVal"] but instead got null`

javivelasco:
<img width="1280" alt="screenshot 2016-02-01 00 29 06" src="https://cloud.githubusercontent.com/assets/1634922/12705958/e0caa426-c87a-11e5-9b9a-b88119a40016.png">

It is for me 😕Which browser are you using?

ipsips:
Safari 9

![screen shot 2016-02-01 at 10 55 58](https://cloud.githubusercontent.com/assets/1770542/12712655/989e72f6-c8d2-11e5-8466-37462a6cbb4b.png)

javivelasco:
Ok, worth investigate it. I'm quite sure it should be a playground thing. Help here appreciated!

ipsips:
cloned the repo and it seems to work just fine locally here

javivelasco:
Ok, thanks! Issue to investigate :)

MIGUELWAXX:
need support.. thanks

goodbeysr:
i copied the code given in the exemple but it doesnt render . i checked the DOM and i found that the tbody is empty even if the source object is full of data

Rabsztok:
@goodbeysr turns out that example on the website is outdated. For example how to use `Table` in `2.0.0-beta`, look at https://github.com/react-toolbox/react-toolbox/tree/dev/components/table#data-table

@javivelasco I think it would be very helpful to mention on top of README, that you should use docs from github source code and not from the website when using 2.0.0.

PR #411, opened 2016-03-26
Update docs

### LINKED ISSUE #405: Update Docs!!!
Related Issues: #124 

I noticed some of the docs are outdated and decided its time for an update. Also gonna look into #124 and see what's causing that issue (no guarantees that I'll be able to fix it but I'll do my best!).

@javivelasco Just wanted to know what's the recommended way to import the react-toolbox components. Look at the options below:

Option 1: ` import { Button } from 'react-toolbox'; `
Option 2: ` import {Button, IconButton} from 'react-toolbox/lib/button';`

KerenChandran:
Fixes #405 

### Description
So far I have updated the docs for all the existing components and fixed spacing (really sorry, its kinda like an OCD). As I write this, I just realized that you might have issues finding what I changed, so if you request to have the docs updated without the spacing I'll go and do that.

### Questions

@javivelasco @felipeleusin 

1. There's a Form Component. I can start writing a doc for it, but should I have it added to the left nav so its visible on the main site?
2. Some of the components which use Font Icon say that the icon value can be `Any` in the proptypes, but it should be `String` or `Element`. Should I update the proptype of the Font Icon Component to be `Any` or update all the other components? The docs in this update match the component's proptypes (e.g. Button will say `Any` for icon and Font Icon will say `String` or `Element` for the value property).

### Other Info
I will look into #124 in the meanwhile and if I find a fix I'll submit it as a separate PR.


felipeleusin:
First of all, wow. Thanks! That was a lot of work. After reviewing this everything seems ok, I'll just wait for @javivelasco to check it out.

I didn't even knew about the Form component. Will look into it as for the FontIcon I think the propType should be node.

javivelasco:
Beautiful! Thanks @KerenChandran 
I'm having some busy days but I'll be able to check it during the next weekend. Just told @felipeleusin by email. Hold on! :)

javivelasco:
Hey @KerenChandran finally checking this! About your questions:

1 - Yeah, there is a `Form` component, but it was not finished yet and I was not sure about the most appropriate approach on it. I know there are already some libraries that are implementing forms so I wanted to give them a look to get ideas. If you think the component is useful as it is :+1: to add it to the docs and index it.
2 - I don't remember the exact reason why we have `any` in the `FontIcon` component but I guess it should be either element or string. @nathanmarks implemented it so probably he has something to say about it!

Great work!! As soon as you get it totally ready I'll be happy to merge

KerenChandran:
@felipeleusin @javivelasco Thanks for your comments. Sorry been busy this week and didn't write the doc for the form component.

- I'll probably index the form component and have a warning saying its experimental and use it at your own risk.
- I'll wait for @nathanmarks and any other comments around 7pm EST (April 6). If there are no objections, I'll convert all the components that use FontIcon to Element or String.

 I'll do my best to get it done tonight so it matches your release schedule.

felipeleusin:
@KerenChandran thanks for your work. I'm honestly not much of a fan of Form components like this, most of the time I use one for more complex forms I end up having to do some nasty hacks and for the simple ones it's easier to just write the form.

Regarding FontIcons my only gripe it's that they are exactly this Font Icons, dependent on loading a custom font. I much prefere SVG icons to font icons, they are easier and MUCH faster than loading an entire font-face and with libraries such as https://www.npmjs.com/package/react-icons they are very simple to use (and you can combine with http://gorangajic.github.io/react-svg-morph/ to some great effects). I was stalling to propose this switch from FontIcons to some sort of SVG icons

KerenChandran:
Looks like I'm done. Also here's some last minute decisions I made.
 
- I decided not to update and index the form component. Judging from your opinions, it sounds like people may be better using something else.
- I also modified the Input's error prop type from node to string as error was defined as a string everywhere else and from a discussion in #434, the error prop is wrapped in a span tag, so it makes sense for it to be a string.
- All components that use the FontIcon Component now check the icon prop type to be a string or element as discussed above.

If there are any other changes that I need to make, please let me know and I'll do my best to get it done asap tomorrow so it doesn't interfere with your release schedule.

javivelasco:
I think it's great. Thanks @KerenChandran !

PR #415, opened 2016-03-27
Fix Snackbar timeout issue

hycner:
This change now wipes previous timeouts when the Snackbar is reactivated

Steps to reproduce the issue:
1) Activate a Snackbar by clicking its Activation Button
2) Reactivate the Snackbar before its timeout activates by either:
    2A) Reactivating the Snackbar by clicking its Activation Button again
    2B) Dismiss the Snackbar via its Dismissal Button, and then click the Activation Button again
3) Observe that the initial timeout will still effect the new activation of the Snackbar

### DIFF
--- a/components/snackbar/Snackbar.jsx
+++ b/components/snackbar/Snackbar.jsx
@@ -18,11 +18,24 @@ class Snackbar extends React.Component {
     type: React.PropTypes.string
   };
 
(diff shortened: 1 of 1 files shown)


javivelasco:
Makes sense!
Thanks :)

PR #416, opened 2016-03-28
Tabs: only render content of current tab

gadicc:
Hey

Currently the contents of all tabs are always rendered, and inactive tabs hidden with `display` block/none.

Is there any reason not to only render the current tab?  Besides being faster, I guess it means potentially that tabs could make use of `componentDidMount()` when they're first rendered (i.e. become the active tab).

This PR is more intended for feedback on the approach.  If you agree, I'll spend a bit more time thinking about any implications and see if it affects any tests, etc.

### DIFF
--- a/components/tabs/Tabs.jsx
+++ b/components/tabs/Tabs.jsx
@@ -82,12 +82,14 @@ class Tabs extends React.Component {
   }
 
   renderContents (contents) {
-    return contents.map((item, idx) => {
-      return React.cloneElement(item, {
-        key: idx,
-        active: this.props.index === idx,
-        tabIndex: idx
-      });
+    const activeIdx = contents.findIndex((item, idx) => {
+      return this.props.index === idx;
+    });
+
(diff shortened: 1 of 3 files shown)


felipeleusin:
Looks good but could you please just check the linting errors?

javivelasco:
Yeah, I think it's a good improvement. If you want to review it a little bit more or just fix linting errors we can merge it for the next release!

gadicc:
Thanks, guys!  I fixed the linting errors, added a test and mentioned the new behaviour in the docs.

Sorry for the delay, was considering whether or not the new behaviour should perhaps be optional, but decided not :)

Also wanted to say that I love the way you have everything structured in the repo, it's a great starting point for others to get involved.

javivelasco:
Thanks for your kind words and your PR!
It will be available in the next release, next weekend.
:)

gadicc:
Thanks guys, for this and the other PRs.

PR #417, opened 2016-03-28
Tabs: ability to (optionally) reference by key instead of index

gadicc:
Hey, this draft PR allows tabs to specify their own key and choose the active tab - optionally - by name instead of the index.  Example usage: the key name is used as part of routing structure; consistent names even when re-ordering tabs; etc.

```jsx
<Tabs tab="fruit">
  <Tab key="fruit" label="Fruit">...</Tab>
</Tabs>
```

This also adds an extra parameter to the callback:

```js
onChange(index, keyNameIfItExists)
```

If you're happy with the idea, I can see if I can make the code any cleaner, add tests, documentation, etc.

### DIFF
--- a/components/tabs/Tabs.jsx
+++ b/components/tabs/Tabs.jsx
@@ -9,6 +9,7 @@ class Tabs extends React.Component {
     className: React.PropTypes.string,
     disableAnimatedBottomBorder: React.PropTypes.bool,
     index: React.PropTypes.number,
+    tab: React.PropTypes.string,
     onChange: React.PropTypes.func
   };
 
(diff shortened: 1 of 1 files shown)


javivelasco:
Yeah, I think it makes sense to have an option like this. Feel free to fix the errors, rebase and don't forget to update the docs!
Thanks :)

gadicc:
Hey, I'm sorry, I've been totally swamped with other work :/  Assuming this isn't dealt with anywhere else, I'll re-open the issue once I've had a chance to do as promised.  Great seeing all the recent commits!

javivelasco:
Great!

PR #420, opened 2016-03-30
docs(media): keep navigation open with playground for wider screens

mgcrea:
The navigation in the docs is a bit painful when you want to quickly play with components in the playground. Having the navigation bar hidden every time you show the playground is burdensome. This fix keeps the navigation open when you have a screen wider than 1440px.

Wanted to automatically switch the playground when you navigate, but encountered issues with `react-router@v1` and mixins+ES6. Could try to upgrade react-router if you want.

### DIFF
--- a/docs/app/components/layout/main/_config.scss
+++ b/docs/app/components/layout/main/_config.scss
@@ -1,5 +1,7 @@
 @import "../../globals";
 
+$media-min-width: 1440px;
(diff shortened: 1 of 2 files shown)


javivelasco:
Sure! I'll have some time this weekend. May you open an issue to track it?
Thanks!!

PR #421, opened 2016-03-31
have Dialog.jsx import from './style' and not './style.scss'

gadicc:
This brings the import in line with other components, and fixes sensitive bundlers that rely on this (i.e. the current early way we're doing it in Meteor).


javivelasco:
My bad! Thanks!

PR #423, opened 2016-03-31
Don't show actions container in dialog if no actions exist

emiloberg:
This PR makes sure that the navigation container of the dialog component isn't shown if no actions exist.

This way we also don't get the (default 0.8em) padded bottom if there's no actions.


TobiasBales:
It would actually be rather nice if you could squash the commits into one.

emiloberg:
Said and done!

javivelasco:
Nice!

PR #426, opened 2016-04-01
Add allowBlank to Dropdown

TobiasBales:
Since @nickw was having issues with the dropdown not defaulting to the
first entry if the value was null (even though the options had a null value in it)
this seems like a good time to add an option to default to the first element of the supplied options
(the prop is called allowBlank) in case the value is not found in the options.

The handling of null/undefined values has been fixed too and if there actually is a value of null
in the options it is correctly selected now.

### DIFF
--- a/components/dropdown/Dropdown.jsx
+++ b/components/dropdown/Dropdown.jsx
@@ -7,6 +7,7 @@ import style from './style';
 
 class Dropdown extends React.Component {
   static propTypes = {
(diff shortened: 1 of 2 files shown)


javivelasco:
Alright! Thanks for the PR. I'm just missing here to add this new prop to the documentation. Once it's done we can merge!

TobiasBales:
Done

javivelasco:
Merged manually!

Issue #430, opened 2016-04-02
Chip component


lucaas:
A component implementing the [Chips specification](https://www.google.com/design/spec/components/chips.html#) would be great.

Ideally, it should be usable both in an Autocomplete component and as a standalone component.

See #174 Autocomplete Enhancements for previous discussion.

Issue #174, opened 2015-11-28
Autocomplete Enhancements


nathanmarks:
This issue is to track enhancements for autocomplete I am currently working on. This includes solving #161, #170  and #138 

I've tried using the autocomplete and found some issues.
1. Doesn't implement the most prominent spec examples. Should focus on providing the user with the tools needed to implement common use cases.
2. Cannot be used easily in complex applications (compatible with a small number of basic use cases).
3. Does too much internally in a single component with little flexibility to show for it

Goals:
- [ ] Solve #161, #170 and #138
- [x] Allow you to connect any type of data source or custom query
- [x] Enable the direct use of `<Input/>` and `<input>` with full access to their component API.
- [x] Can be easily used for everything in the spec such as the raised boxes, search bars, appbar search fields, full width searches on mobile when responsive styling is implemented.
- [x] Support for autocomplete item templating.
- [ ] Still have the option to use chips, but with greater control.

This means that all input related styling/settings props can be passed to that component as described in the documentation. That way the autocomplete implementation is also more portable and focuses on providing the elements unique to autocomplete.

javivelasco:
Awesome @nathanmarks

soyjavi:
Great!!

kagux:
I've came across multiple situations where I need to customize selection list. In fact, because we use Flux, it would be much easier to have autocomplete input separated from selection. We ended up rendering custom selections using data from stores, not from autocomplete component. 

I wish there were 3 components: Autocomplete, Chip and maybe MultiSelect to combine them. That way I could take Autocomplete, connect it to store and use a List to render selection.

Anyways, @nathanmarks do you still have time to work on this issue? Is there any way I could help?

nathanmarks:
@kagux Yeah I already started on this and actually just built a react autocomplete out the other day for work that uses redux that I intend to port some of the functionality from.

acornejo:
@nathanmarks do you have a branch that you would like to share? I've run into some problems with autocomplete myself, and I would rather reuse someone elses' work instead of reimplementing it.

javivelasco:
I'm probably working on this too in a week or so. It would be cool to open a branch here so more people can contribute

nathanmarks:
@javivelasco @acornejo I will definitely try to push a branch then this week! I just want to clean some things up first before passing the baton.

javivelasco:
:+1: !

nathanmarks:
@javivelasco @acornejo 

[...]

PR #431, opened 2016-04-02
Add chip component

lucaas:
Adds a basic chip component implementing the [Chips specification](https://www.google.com/design/spec/components/chips.html#) to address issue #430.

Some things to consider:

* The delete icon is larger than in the reference images, but follows the spec sizes.
* I didn't add a explicit `onDelete` property, but simply rely on the standard `onClick`.
* Which approach would be more in line with the rest of the components, passing in `avatar` and `label` as props, or child components?
* I've replaced the rendering of autocomplete values with chips, but left it there as I wanted to keep this pull request small and not tackle the complete scope in #174.

Please let me know if there are anything more I should address.


acornejo:
This is great Lucaas! I hope the maintainers take a look at this PR soon.

I have a few questions though, would the onClick allow a user of this component to distinguish between someone clicking the chip and someone clicking the delete button? Both events seem like could be of interest.

Also it would be nice if in addition to being able to specify a label (which is what most use cases want), one could instead pass children components to do custom labeling (for instance, to display user name where firstname is bold and lastname is not, etc..).

In any case, any version of this component seems like a great addition to react-toolbox! Thanks!

javivelasco:
Hi! Thanks for this new component! As @acornejo pointed, I think it should be possible to give raw children but that's not a change that should take too long to do. Apart from that, I cloned the branch to check the changes and noticed it can be just a little tweaked to look even better.

I had an approach for the Chip in local that never updated so I change compare implementations a maybe fix this small pixel-level details. I'll do it during the day a release a new patch so you guys can start using it now!

lucaas:
Thanks for the feedback!

I can take a stab at adding a separate `onDeleteClick` prop and moving `label` and `avatar` to children instead.

Styling it was a bit trickier than I expected, you are more than welcome to update it.

- [ ] Add `onDeleteClick` property
- [ ] Use children instead of special `label` and `avatar` props in Chip component.

lucaas:
Done.

Please check that you agree with how I check if the component has an `Avatar` child. I am not sure if that is best practice, but need a way to control the padding of the chip and sizing of avatar if one is present.

On a related note, I am not entirely sure about the `deletable` prop either, would it better to rely on the presence of a `onDeleteClick` callback? Or should it be named something else to make it more intuitive, maybe `readonly` (with inverse behaviour)?

When I added the `onDeleteCallback`, I also removed the hover and active styles for the chip since I imagine that they wouldn't commonly be wanted.

kagux:
great work @lucaas !
would it be possible to add animated or static gif of how it looks in PR?

lucaas:
Thanks @kagux. I haven't added any animations, so a static image should be enough; Here are the autocomplete and chip examples from the spec page:

screen shot 2016-04-04 at 12 11 03

screen shot 2016-04-04 at 12 11 30

lucaas:
After looking at the spec page again, I think I confused the size of the "remove area" and filled circle with the remove icon. It appears as it doesn't fill the whole area, and will probably look better as a smaller circle.

- [ ] Make remove icon smaller.

I also thought about if the use of `FontIcon` for the remove icon was appropriate, or if it would be better to use a more self-contained inline SVG to avoid having a dependency on the Material design icon font for something so trivial.

acornejo:
+1 for your inline SVG idea. However, if that is something that will take you a while to do, I think its better to get *some* version of your PR merged and it can be polished later. Also from your screenshot it seems clear the delete icon is not vertically centered in the chip (its closer to the bottom than it is to the top), but again I think that is a problem we can live with for the first version of this component.

great work!

lucaas:
I've pushed a new commit which replaces the FontIcon with an inline SVG and updates sizes to better match how it looks in Google's spec. 

screen shot 2016-04-05 at 10 52 51

javivelasco:
Really beautiful @lucaas !! Congrats and thanks for the great work! 💃🏻

aaj:
Not sure if I broke something, but I'm on 1.2.1 and my chips look like this:

image

acornejo:
its definitely something wrong with your setup. I have an app running on 1.2.1 using chips and it looks just fine. Based on your screenshot even the font looks wrong, probably not importing your css correctly (i.e. incorrect webpack config).

I suggest you try to reproduce your problem with chips on this project https://github.com/react-toolbox/react-toolbox-example

javivelasco:
Maube you are missing the commons.css imports or the Roboto font?

PR #433, opened 2016-04-05
Add prop to autocomplete to place labels below or above input

ro-savage:
I have updated the `Autocomplete` component to take a `selectedPosition` prop which allows you to choose if the selected labels should go above or below the input.

Defaults to `above`. 
Accepts `above` or `below`.

Also updated documentation  (Am I missing somewhere else?).

Happy to make changes if there is a better way to do this, or if I need to update extra documentation or updated something wrong (Not sure how react-toolbox.d.ts works?)

### DIFF
--- a/components/autocomplete/Autocomplete.jsx
+++ b/components/autocomplete/Autocomplete.jsx
@@ -11,10 +11,16 @@ const POSITION = {
  UP: 'up'
 };
 
+const SELECTEDPOSITION = {
+  ABOVE: 'above',
+  BELOW: 'below'
+};
+
 class Autocomplete extends React.Component {
  static propTypes = {
    className: React.PropTypes.string,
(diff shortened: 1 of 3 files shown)


javivelasco:
I think this is ok but as something temporary thought. I think the best solution would be to remove the autocomplete selection from the `Autocomplete` component. The selected fields should probably be outside the component so that a parent would be able to render `Chips` or other component.

I have no time to do this right now, but I think it would be a great improvement. I'm going to merge this anyway while the split is coming.

Thanks for your contribution!

paddotk:
This seems to conflict with the `direction` prop but in no case does what it's supposed to do (played with it on react-toolbox.io). Am I understanding its intended behaviour wrong or did it bug out?

PR #444, opened 2016-04-10
Render on demand

javivelasco:
There are some components that have an `active` property that indicates if it should be shown. An example is the `Dialog` and the `Drawer`. Since they are animated via CSS when the become active, they are rendered since they are present in your JSX. This means that if you have 10 dialogs in the same view, all the content is rendered even though is not used.

This should be easy to avoid just avoiding to render when the component is active but this also means that when it becomes active it cannot be animated since the transition does not happen. It is directly rendered with the classes that should produce the animation.

This PR adds a new HOC for components that can become active. It accepts and `active` and when it's received to `false` nothing is rendered. When the prop changes to `true`, instead of passing the prop first it's going to render the `children` passing `false` and then, after a 20ms window, it's going to re-render with the prop set to `true`. This means a `Dialog` for example can be animated when it becomes active. When it becomes inactive though, `children` cannot be un-rendered directly because probably an animation should happen. Because of this we considered a time window of `500ms` by default before deleting the content from the DOM.

cc @kagux @ro-savage 

Also, I think the `App` component is not needed anymore 💃🏻


ro-savage:
Looks good @javivelasco
The ActivableRenderer.js is pretty much word-for-word the HoC I made to handle the issue in my project. Should have just done a PR ;)

Will be great when this is released and I can remove the HoC out of my project.

Thanks for your awesome work as usual!

javivelasco:
Thanks @ro-savage it's already released! :)

PR #447, opened 2016-04-12
Table: ability to select only one row at the time

mjurincic:
In our app we needed the ability to let our user select just one option from list of choices presented in table, the way the table works now either all rows are selectable if you make selectable true (by default its always true) or none if you specify false, another bool prop multiSelectable was added to account for that use case and is by default true so the table will behave as before unless you provide 
multiSelectable = {false}

Don't know if this is proper way to handle this so I'm opening this draft PR to start discussion, If there is interest from other people to have this functionality, I can see if I can make the code any cleaner, add tests, documentation, etc.

### DIFF
--- a/components/table/Table.js
+++ b/components/table/Table.js
@@ -11,6 +11,7 @@ class Table extends React.Component {
     onChange: React.PropTypes.func,
     onSelect: React.PropTypes.func,
     selectable: React.PropTypes.bool,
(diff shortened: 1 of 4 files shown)


javivelasco:
Retest this please

javivelasco:
Thanks!

Issue #448, opened 2016-04-12
body  overflow:scroll issue on new Portal.js HOC Overlay


joevbruno:
In the new HOC (react-toolbox/components/hoc/Portal.js), the unrenderOverlay method causes a horizontal scrollbar to appear unnecessarily in at least Google Chrome. Furthermore, because it appears as an inline style, it's specificy wins. In the code below, document.body.style.overflow needs to be set to null instead of 'scroll'. I am happy to submit this fix as a pull request. Am I missing something that requires the body style to be set to scroll?

```
 _unrenderOverlay () {
    if (this._overlayTarget) {
      if (this.props.lockBody) document.body.style.overflow = 'scroll';
      ReactDOM.unmountComponentAtNode(this._overlayTarget);
      this._overlayInstance = null;
    }
  }
```

javivelasco:
I missed that, both the x-overflow and the property value. It should be at `auto` instead of `scroll` when it's re-enabled. I'll be happy to accept a PR about it and release a patch :)

dlebedynskyi:
have same issue. fix would be awesome )

jasonkylefrank:
I work with @joevbruno and have discussed this with him... along with the `overflow = 'null'` PR that he submitted, we think we know of a way to let users get the behavior that they need.

I checked Joe's PR and although it partially addresses our need (which is for React Toolbox to not mess with our `body` styles), it has some unwanted side effects such as delayed removal of the scrollbar - making the opening / closing of the `Drawer` not as smooth as it was (because the space for the vertical scrollbar that appears/disappears on Windows causes the `Drawer` to "jump" during that entry/exit because of the `setTimeout` - which is not synched to the `Drawer` opening/closing).

Therefore, along with his PR, a better option would be to let users "opt-out" of your control messing with the `body` style... and the great thing is that you already have most of the hooks/infrastructure in place to provide this!

Just provide a prop for `Drawer` for `lockBody` and pass that down to `Portal`.  `Portal` already checks for this prop, but we have no way to pass it in...

![image](https://cloud.githubusercontent.com/assets/7502365/14473110/8a4b2f58-00c4-11e6-8afe-37aa03df44bb.png)

![image](https://cloud.githubusercontent.com/assets/7502365/14473132/a1a8b346-00c4-11e6-88b5-82657b904b51.png)

[...]

PR #449, opened 2016-04-12
issue: 448 body scroll in HOC Portal.js

joevbruno:
Submitting a PR to fix issue #448 where document.body.style.overflow was set to scroll, causing a horizontal scroll bar to appear unnecessarily. Changing the value to auto does not fix the issue. Because 'auto' is still an inline style, it will also override any CSS styles on the body (such as overflow-x: hidden). We should allow developers to handle the CSS styles on the body as they desire. This can be accomplished by setting document.body.style.overflow = null (which removes it from the DOM) if toolbox does not have an explicit need to scroll.

### DIFF
--- a/components/hoc/Portal.js
+++ b/components/hoc/Portal.js
@@ -68,7 +68,7 @@ class Portal extends Component {
 
   _unrenderOverlay () {
     if (this._overlayTarget) {
-      if (this.props.lockBody) document.body.style.overflow = 'scroll';
+      if (this.props.lockBody) document.body.style.overflow = 'auto';
       ReactDOM.unmountComponentAtNode(this._overlayTarget);
       this._overlayInstance = null;
     }
--- a/components/hoc/Portal.js
+++ b/components/hoc/Portal.js
@@ -68,7 +68,7 @@ class Portal extends Component {
 
   _unrenderOverlay () {
     if (this._overlayTarget) {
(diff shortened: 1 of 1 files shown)


components/hoc/Portal.js
```diff
@@ -68,7 +68,7 @@ class Portal extends Component {
 
   _unrenderOverlay () {
     if (this._overlayTarget) {
-      if (this.props.lockBody) document.body.style.overflow = 'scroll';
+      if (this.props.lockBody) document.body.style.overflow = null;
```
javivelasco: Is this the same as setting `auto` which I think it's the default?

joevbruno: The problem is because it is an inline style, 'auto' will override any CSS. For instance, I have overflow-x: hidden set on my body. Setting document.body.style.overflow = 'auto' overrides my CSS, and there is no way for me to override react-toolbox's inline style (because inline always wins).

joevbruno: Updated comment.

javivelasco:
I left some thoughts! Check #448 ! :)

ro-savage:
I created PR #470 to address the same issue in a different way.

javivelasco:
It's already fixed https://github.com/react-toolbox/react-toolbox/commit/8a7ddc7b96f14a87b8f1e5f4e8ed07ceb79a13ae Thank u guys for this!

PR #452, opened 2016-04-13
Fix the value of month in datepicker example

magicae:
The value `date.getMonth()` returns is from 0 to 11.


javivelasco:
Thanks! :)

PR #457, opened 2016-04-15
Cleanup timers when Menu component unmounts to prevent exceptions

mattkauffman23:
I'm using Menu components inside an infinite scroll grid where non visible rows get unmounted. When users scroll quickly the Menu components can get unmounted before their timeouts fire, throwing exceptions when the timeout functions does eventually get invoked. This change stores handles to the timeouts and clears them when Menu unmounts.


javivelasco:
Thanks!

PR #462, opened 2016-04-19
update check empty actions

Lupino:
```javascript
<ListItem>
    <span> this may no render without avatar or icons </span>
</ListItem>
```

reason is that `emptyActions` do not check the third item fixed follow

```javascript
const emptyActions = (item) => !item[0] && !item[1] && !item[2];
```


javivelasco:
Thanks!

PR #463, opened 2016-04-20
re #448: add / remove noScroll class instead of settig inline style directly

emirotin:
This has conflicts as I forked my branch from `master` to run our own tests on the stable codebase. If you approve the idea I will rebase it against `dev`.

I've also renamed the phantom dependency name to suppress the warning.

### DIFF
--- a/components/hoc/Portal.js
+++ b/components/hoc/Portal.js
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import ReactDOM from 'react-dom';
+import style from './style';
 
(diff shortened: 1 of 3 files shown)


javivelasco:
Actually I've changed this but it's not pushed yet. I think the approach should be to lock on the overlay instead of the portal. This way the lock happens without taking into account the delay window used to perform the animation. Gonna push today so you can check it!

emirotin:
Nice!

javivelasco:
It's already fixed. About the lib thing, we are going to include it!

Issue #466, opened 2016-04-21
"Module build failed"


amensouissi:
Hello,
I get this error when try to start the project...

npm install
npm start
...
Child extract-text-webpack-plugin:
    
    ERROR in ./~/css-loader?sourceMap&modules&importLoaders=1&localIdentName=[name]__[local]___[hash:base64:5]!./~/postcss-loader!./~/sass-loader?sourceMap!./components/snackbar/style.scss
    Module build failed: 
      @import "./colors";
     ^
          Import directives may not be used within control directives or mixins.
          in /home/amensouissi/workspace_react/react-toolbox/components/_base.scss (line 4, column 3)
....

digipi:
Having the same issue. I'm using the react-redux-starter-lot by Dave Zuko. Any ideas?

vincentfretin:
removing the if in _base.scss and just having
````
@import "./colors";
@import "./globals";
@import "./mixins";
```
fixes the issue.

amensouissi:
#451

dlebedynskyi:
can you please make patch release for this? CI is going nuts due to this error

nickretallack:
Yeah, it'd be nice to have a working package that can be installed without modification.  Could you resolve this somehow?

You could apply this change to core, though I'm not sure how you'd reconcile this with your theming system.

Or, you could specify a version dependency for sass that prevents it from upgrading to the version that breaks it.

nickretallack:
Oh, huh, it is specified to use 3.4.2 but it still breaks for me.  Perhaps it's a dependency of a dependency that is at fault.

dlebedynskyi:
note this 
https://github.com/sass/node-sass/issues/1459#issuecomment-212839565

this is a bug. won't fix. might need to fix lib  here in first place to be compatible

nickretallack:
There's a few other syntax errors too, it seems, at least [as far as the linter is concerned](https://travis-ci.org/react-toolbox/react-toolbox/builds/124861860).  I'm not sure what's wrong with these lines, though.

[...]

PR #469, opened 2016-04-21
Remove conditional import since SASS does not support them.

nickretallack:
Based on @vincentfretin's comment on issue #466.


javivelasco:
This would break customization via loader I think but it is ok since for the next release the customization it's going to be different!

PR #471, opened 2016-04-22
Fixed sass-lint to 1.5.1 to avoid linting errors

ro-savage:
See Issue #472


ro-savage:
Fails due to SASS import statement. Fixed by PR #469

Issue #473, opened 2016-04-22
DatePicker doesn't handle being passed an empty string


ro-savage:
React 15 throws warning when passing an input `null` and prefers an empty string.

The DatePicker throws propType errors if it receives an empty string, and doesn't handle it well in general.

> Future versions of React will treat <input value={null}> as a request to clear the input. However, React 0.14 has been ignoring value={null}. React 15 warns you on a null input value and offers you to clarify your intention. To fix the warning, you may explicitly pass an empty string to clear a controlled input, or pass undefined to make the input uncontrolled.

ro-savage:
PR #474 fixes this issue.

PR #474, opened 2016-04-22
Update DatePicker to handle being passed a value of an empty stirng

ro-savage:
This issues issue #473 that React 15 now throws warning when passed null, and prefers empty strings.

DatePicker should be able to handle being passed an empty string.

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -17,7 +17,10 @@ class DatePicker extends React.Component {
     maxDate: React.PropTypes.object,
     minDate: React.PropTypes.object,
     onChange: React.PropTypes.func,
-    value: React.PropTypes.object
+    value: React.PropTypes.oneOfType([
+      React.PropTypes.instanceOf(Date),
+      React.PropTypes.string
+    ])
   };
 
   state = {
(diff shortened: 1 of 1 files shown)


components/date_picker/DatePicker.js
```diff
@@ -17,7 +17,10 @@ class DatePicker extends React.Component {
     maxDate: React.PropTypes.object,
     minDate: React.PropTypes.object,
     onChange: React.PropTypes.func,
-    value: React.PropTypes.object
+    value: React.PropTypes.oneOfType([
+      React.PropTypes.instanceOf(Date),
```
ro-savage: Changed to checking for a Date object, instead of just an object.

components/date_picker/DatePicker.js
```diff
@@ -41,7 +44,8 @@ class DatePicker extends React.Component {
   render () {
     const { inputClassName, value } = this.props;
     const inputFormat = this.props.inputFormat || time.formatDate;
-    const date = value ? inputFormat(value) : null;
+    const date = Object.prototype.toString.call(value) === '[object Date]' ? value : undefined;
```
ro-savage: Check if the value is a date or something else.

javivelasco:
Great work! Thanks! :)

Issue #445, opened 2016-04-11
Datepicker dialog and input date get out of sync


Siern:
When the value property is dynamically changed for the date picker, the DatePickerDialog still retains its old date.
This appears to be because the Dialog is re-rendered rather than recreated. 
Due to this code and its reference for the dialog:
```
state = {
    date: this.props.value,
    display: 'months'
  };
```
```
          <div className={style.wrapper}>
            <Calendar
              display={this.state.display}
              maxDate={this.props.maxDate}
              minDate={this.props.minDate}
              onChange={this.handleCalendarChange}
              selectedDate={this.state.date} />
          </div>
```

Now in the DatePicker's render
```
        <DatePickerDialog
          autoOk={this.props.autoOk}
          active={this.state.active}
          className={this.props.className}
          maxDate={this.props.maxDate}
          minDate={this.props.minDate}
          onDismiss={this.handleDismiss}
          onSelect={this.handleSelect}
          value={this.props.value}
        />
```
This props.value sent to the dialog on render is not used! Instead the dialog reverts to the old state value.

I ran into this issue when putting the datepicker into a compound control where a select was used to choose from stock date entries.  So that when a selection was made it changed the datepicker value.
These changes were fed back to the current view via property stream.  Hence this datepicker simply had its date value changed in the model and the properties streamed reflected this new value.  The input properly updates and the dialog it opens it still on the original date from the initial render of the page.

ro-savage:
I have created a PR #475 to fix this issue.

eagsalazar:
At this point this component has a critical bug but it isn't really clear what the correct behavior is.  It is pretty strange for the props to change while the dialog is open but it does it seems fair to just reset the state in which case something quite simple would work wouldn't it?

```
  componentWillReceiveProps (nextProps) {
    if(!this.props.active && nextProps.active || this.props.value !== nextProps.value) {
      this.setState({ displayTime: nextProps.value })
    }
  }
```

Siern:
Well the problem I am describing is that for some reason state is being stored in the control and only being refreshed if the control is recreated.

I would patch it by respecting the property for the selected date that was received by the render request rather than ignoring it in the dialog render.  As I would expect the control to act as stateless (data-wise) and rely on the model to feed it, in my case via property streams.

[...]

PR #475, opened 2016-04-22
Fix DatePickerDialog to correctly display the previously selected date

ro-savage:
This fixed issue #445

The calendar was displaying todays date every time it was opened, rather than the date contained in the input field.

### DIFF
--- a/components/date_picker/DatePickerDialog.js
+++ b/components/date_picker/DatePickerDialog.js
@@ -24,9 +24,18 @@ class CalendarDialog extends React.Component {
   };
 
   state = {
-    date: this.props.value,
-    display: 'months'
+    display: 'months',
+    date: this.props.value
   };
+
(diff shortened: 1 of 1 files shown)


components/date_picker/DatePickerDialog.js
```diff
@@ -23,10 +23,22 @@ class CalendarDialog extends React.Component {
     value: new Date()
   };
 
-  state = {
-    date: this.props.value,
-    display: 'months'
-  };
+  constructor (props) {
```
javivelasco: May you use class properties for this? Check the other components!

ro-savage: @javivelasco Fixed up. Just automatically used constructor without thinking :)

javivelasco:
Thanks!

Issue #82, opened 2015-11-09
JSPM


giltig:
Hi,
right now there's only support for webpack.
using your great library with jspm is very inconvenient because you need to build the library all the time instead of using it raw.
Can you please add support for jspm

javivelasco:
Hi, thanks for your words :)

I would like to support other module loaders such as JSPM or browserify since there is already an example implementation with CSS modules. I never used JSPM before so help here is really appreciated. Let's see if somebody can help with it.

Thanks!

shidhincr:
@javivelasco I would like to pick this up. Let me try how I can do this.

javivelasco:
Thanks sounds awesome! Thanks!! :)

rockneverdies55:
+1 for browserify integration...

bali182:
+1 for jspm! I started learning it around a week ago, but if nobody started it, I will give the jspm integration a shot, because this looks like THE material library for react, and it would be a shame to leave out poor jspm ;) Btw guys, really good job, it's rare, that I look at a library, and the structure & code just looks clear.

bali182:
There will be problems with the sass. The best looking sass-plugin I found is this: https://github.com/mobilexag/plugin-sass , but it requires to load the `.scss` files like this: `import './style.scss!'` (notice the `.scss!`). And even when doing `import style from './style.scss!'` it just appends the contents of the compiled `.scss` to the documents head, and the `style` object is empty. This needs some serious work.

shidhincr:
@bali182 I was trying to implement the JSPM version, but as you said the sass files loading is pretty difficult. JSPM requires non-js files to be imported with `!` at the end of the file name.

I was thinking about keeping a separate build for JSPM, by re-writing these imports using a gulp task. I am not sure if that's the right approach.

bali182:
Damn, sorry, looking back I see now, that you are working on it as well. Last night I was going through how webpack is doing this, and how jspm is, but it's still a bit unclear, if the `./filename.extension!` style import is hardcoded in jspm for non js imports? Or is there a chance, that we could hack together a custom sass loader plugin, which is capable of handling the existing import formats? That sounds (in my opinion at least) a bit safer, than doing the import replaces in gulp. Maybe I'm wrong, since that would require to somehow take over the loading if for example `./style.js` is not found...

shidhincr:
[...]

PR #477, opened 2016-04-22
Fix for jspm

etnbrd:
#82 

jspm load non js files with special loaders. For sass, it uses [plugin-sass](https://github.com/mobilexag/plugin-sass).
These special loaders are invoked by the presence of the `!` character, like this : `import './style.scss!';`.

Without this, jspm expects a simple .js file, therefore, it raises exceptions when it fails to import `style.js`.
To resolve this issue, I simply added a `style.js` file that load the corresponding scss file in each components directory.

As the files holds the scss extension, jspm would look for a scss module, not a sass module.
To correct this, sass needs to be specified at the end of the import, like this : 

```
import './style.scss!sass';
```
These files are then compiled and put directly into the lib folder.
And apparently it works :)

### DIFF
--- a/components/animations/style.js
+++ b/components/animations/style.js
@@ -1,0 +1,1 @@
+import './style.scss!';
--- a/components/animations/style.js
+++ b/components/animations/style.js
@@ -1,1 +1,1 @@
-import './style.scss!';
+import './style.scss!sass';
--- a/components/app_bar/style.js
+++ b/components/app_bar/style.js
@@ -1,0 +1,1 @@
+import './style.scss!';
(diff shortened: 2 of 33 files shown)


javivelasco:
Does this work for webpack as well? I think there should be a way to set a context. Via env variable or whatever to add this customizations. We can think further about it during the weekend, I'm going to PR something that makes tooltox potentially usable without webpack

javivelasco:
Closin! I think this is not going to be needed with the new style approach, thanks :)

cpmech:
Hi, I'm just wondering if react-toolbox is finally working with jspm? (I've tried quickly and is not working). Many thanks!

javivelasco:
It should!

PR #478, opened 2016-04-25
Fixes typo in ListItemLayout

justinrainbow:
If no "leftActions" were defined on a ListItem, the "rightActions" will
not be rendered due to a typo.


javivelasco:
Thanks! 😅

binhqd:
I got the same issue.
When will it have new release with this bug fix?

PR #480, opened 2016-04-27
Fix #454: Chip component unable to display nicely on small screen

### LINKED ISSUE #454: chip component unable to display nicely on small screen
http://react-toolbox.com/#/components/chip  try adjust your browser and make it smaller, each of the chip is not aligning nicely

lucaas:
Added text truncation to chip component and limit its size to the the size of the containing element.

screen shot 2016-04-27 at 11 53 52

Fixes #454: Chip component unable to display nicely on small screen

### DIFF
--- a/components/chip/style.scss
+++ b/components/chip/style.scss
@@ -3,12 +3,17 @@
 @import "./config";
 
 .chip {
+  position: relative;
   display: inline-block;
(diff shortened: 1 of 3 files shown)


javivelasco:
Makes sense, thanks!

PR #482, opened 2016-04-27
Fix #481: Body text line height is too large

### LINKED ISSUE #481: Body text line height is too large
I thought the line height of body text (paragraph elements) was too large, so I modified it my application code. When doing that, I looked at the [spec](https://www.google.com/design/spec/style/typography.html#typography-styles) and it seems to specify the standard body text leading as 20pt, while react-toolbox specifies it as 24pt (2.4 units).

lucaas:
Adjust body text line height to conform to material design spec, fixes #481.

The [spec](https://www.google.com/design/spec/style/typography.html#typography-line-height) specifies the leading as 20pt, which should correspond to a line height of 2.0rem in react-toolbox.

Please note that this change may have unwanted side effects as sizes of paragraph elements will change. I have not tested the implications on all components.

### DIFF
--- a/components/_mixins.scss
+++ b/components/_mixins.scss
@@ -120,7 +120,7 @@
   @include typo-preferred-font($use-preferred);
   font-size: 1.4rem;
   font-weight: 400;
-  line-height: 2.4rem;
+  line-height: 2rem;
   letter-spacing: 0;
 
   @if $color-contrast {


lucaas:
Comparing the rendered output, I am not sure that changes are enough. Material design, and react-toolbox, defines two sizes for body type, *Body 1* and *Body 2*.

* Body 1 is used for <p> elements in `commons.scss`.
* Body 2 is not used, except for CardTitle's with avatars.

Since there was no difference in react toolbox between the two previously, all *Body 1* appeared as *Body 2*. This change makes everything appear as *Body 1*, even though certain components should still appear *Body 2*.

Example, card text:

Spec:

screen shot 2016-04-27 at 13 52 35

Before:

screen shot 2016-04-27 at 13 50 18

After:

screen shot 2016-04-27 at 13 49 33

javivelasco:
I'm not sure about this. I'll try to dig deeper during the weekend but those mixins are taken from material design lite so they should be accurate to the spec :\

lucaas:
There seems to be some differences between different projects. *Material Design Lite* specifies the line height as 24px ([source](https://github.com/google/material-design-lite/blob/master/src/_mixins.scss#L135)), while *Angular Material* specifies it as 2rem ([source](https://github.com/angular/material/blob/master/src/core/style/mixins.scss#L71)) and *MUI* as 20px ([source](https://github.com/muicss/mui/blob/cec9cfe1ab064cc2a7216eafa9c8b65a2555df87/packages/npm/lib/sass/mui/mixins/_typography.scss#L58)).

javivelasco:
I'd like to keep it as material design lite. If somebody come to this PR thinking the line height is too big, please reopen, we can discuss further :) Thanks @lucaas

PR #484, opened 2016-04-27
Fix #483: Autocomplete unintuitive filter behavior with multiple=false

### LINKED ISSUE #483: Autocomplete unintuitive filter behavior with multiple=false
The Autocomplete component with multiple=false behaves unintuitively after a value has been set (in my opinion, at least).

Once a value has been set, the dropdown will only show the selected item, since that is the only match. To change the value, the user must first manually clear the input field before he can select something else. I think it would be a better user experience to show all values until the query has been changed.

For reference on the desired behaviour, see the [react select](http://jedwatson.github.io/react-select/) examples for single selection. 

An alternative approach is taken by [angular material](https://material.angularjs.org/latest/demo/autocomplete), which includes a clear input icon, and does not show suggestions when a value has been set.

lucaas:
Change autocomplete filter behaviour when multiple=false

Adds a showAllSuggestions state, which is set to true initially.
It is set to false when the input is changed, and is used to control if
the shown suggestions should be filtered based when the multiple prop is false.
Once a value is set, the state will be reset to true.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -127,18 +127,32 @@ class Autocomplete extends React.Component {
  }
 
  suggestions () {
-   const suggest = new Map();
+   let suggest = new Map();
    const query = this.state.query.toLowerCase().trim() || '';
    const values = this.values();
-   for (const [key, value] of this.source()) {
-     const valueMatchesQuery = value.toLowerCase().trim().startsWith(query);
-     if (
-       (valueMatchesQuery && !values.has(key))
-       || (!this.props.multiple && (this.state.showAllSuggestions || valueMatchesQuery))
-     ) {
-       suggest.set(key, value);
+   const source = this.source();
+
+   // Suggest any non-set value which matches the query
+   if (this.props.multiple) {
+     for (const [key, value] of source) {
+       if (!values.has(key) && value.toLowerCase().trim().startsWith(query)) {
+         suggest.set(key, value);
+       }
      }
+
+   // When multiple is false, suggest any value which matches the query if showAllSuggestions is false
+   } else if (query && !this.state.showAllSuggestions) {
+     for (const [key, value] of source) {
+       if (value.toLowerCase().trim().startsWith(query)) {
+         suggest.set(key, value);
+       }
+     }
+
+   // When multiple is false, suggest all values when showAllSuggestions is true
+   } else {
+     suggest = source;
    }
+
    return suggest;
  }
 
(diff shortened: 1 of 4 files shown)


javivelasco:
Maybe this should be initialized via prop?

lucaas:
Maybe so, yeah. Do you imagine the current behavior would be wanted? Which behavior would you prefer to be default? Do you have any suggestions for the prop name?

Also, I just noticed that @grammka already had addressed the same issue in a very similar, but slightly different way in PR #455. It would be good to compare our solutions before merging.

grammka:
I rly can't imagine when users need current behavior.. If u can give example plz. My thoughts about this: for example we selecting "My Country". I selected Russia and then I want to select other one, so I have to clear field and then type smth else or select from list.. Look at [selectize](http://selectize.github.io/selectize.js/) - there is **Single Item Select** section.

lucaas:
I brought up two other libraries in #483. React select, which behaves similar to selectize and Angular Material, which does not show any suggestions until the input is changed and provides a clear icon. 

Going through old issues, this was also brought up in #252:

> When we have some selected option, Then user want to change this. Now he need to remove all text from input, then choose onother option from list. I think the best option is: 
-Remove text from input, and show all options when user click on input (like a simple select)

> I think the best option, is add one props to component which describe that clean input on click or not, and show all options in list when multiple = false.

grammka:
> I think the best option, is add one props to component which describe that clean input on click or not, and show all options in list when multiple = false.

Sure, It's good way. So how to be with my PR?)

lucaas:
I compared the solutions in a4810d8 and 77190b8, and the differences are:

* 77190b8 has added logic to clear the input on backspace, if the field has not been changed.
* 77190b8 resets the state on query blur and value change, a4810d8 only resets it on value change. What is the rationale for resetting it on query blur?
* The `suggestions()` method has been refactored differently, but I believe the output is the same.

I thought some more about adding the prop, but couldn't really think of any intuitive way of explaining the different behaviours with a short name, so I'll hold off for now.

grammka:
@lucaas I like ur solution with reseting state prop on queryChange, I used other way, more complex - set prop to **true** on focus and revert it on close(blur) / select..)) I think ur code much better ;) Then I'd like u add logic to clear the input on backspace.

I don't know about prop name.. but I have other idea - mbe it's better to add "select" component with such logic?

KerenChandran:
Ya so I created the fix with #399 cause the selected Item wasn't rendered even if a valid option is already selected. The way its implemented, it acts like a google search field component, where it filters it down and as you delete it, you'll get more options.

I do like the suggestion of populating the other entries of showing all the elements once an element is selected. I think we can do that with a prop, but I'm concerned what effect it will have on the multiple={true} type autocomplete list. I can probably look into this Sunday or sometime next weekend.

lucaas:
I've pushed some new commits with the following changes:

* Add prop, `showSuggestionsWhenValueIsSet`, to control the behaviour.
* Clear the query input on backspace, when a value is selected and the behaviour is true.
* Refactor the suggestions method to make the different code paths more understandable
* Add a new autocomplete to the spec page to show the two different behaviours.

PR #489, opened 2016-04-28
Hotfix for incorrect ternary statement in DatePicker

ro-savage:
This fixes issues with PR #474 where the formatted date ternary's true/false was wrong way round.

Sorry about that @javivelasco, I must have switched !== to === after I finished testing, and not swapped the true/false values.


javivelasco:
Thanks! :)

PR #492, opened 2016-04-28
Dropdown with icon

jyoona:
Added the ability to add icon to dropdown fields

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -3,6 +3,7 @@ import ReactDOM from 'react-dom';
 import ClassNames from 'classnames';
 import Input from '../input';
 import events from '../utils/events';
+import FontIcon from '../font_icon';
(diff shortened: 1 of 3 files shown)


hamzapurra:
I don't think that is needed. It already works for Dropdown. Dropdown passes it via others to Input.

jyoona:
Do you mean template? Or is icon actually already a thing?
If it's for template we need to add an icon to each dropdown element, which is different from what I'm trying to implement.

javivelasco:
@hamzapurra is right. [Here](https://github.com/react-toolbox/react-toolbox/blob/master/components/dropdown/Dropdown.js#L133) additional props are passed down to `Input` and it accepts an Icon. So it should work already

javivelasco:
I'm going to close to avoid noise. In case you can't make it work, please reopen.
Thanks for your pr!

PR #470, opened 2016-04-22
Fix issue with body overflow in HOC portal.js

ro-savage:
This fixes issue #448 where the overlay was applying the style `overflow: scroll` to the `body` after being closed.

PR #497, opened 2016-05-01
fixed overflow for unrender overlay

Formatc1:
I fixed adding style to element body from `overflow: scroll` to `overflow: auto` when overlay is hidding. Now it doesn't add inactive scrollbars to pages with content fitted in window.

Second commit fixed colors in button accent-contrast style.


ro-savage:
PR #470 addresses the same issue, but instead checks the previous value of property and re-adds it, or remove the property entirely if there wasn't a overflow to start with.

Hasn't been merged. So maybe @javivelasco wants to just got with a more straight forward solution as presented here.

javivelasco:
Guys, thanks for the PRs. I thought I've mention a solution already but I forgot about pushing it. It [here](https://github.com/react-toolbox/react-toolbox/commit/8a7ddc7b96f14a87b8f1e5f4e8ed07ceb79a13ae). It's solving the same thing so I'm closing!

PR #502, opened 2016-05-05
Add tabs on the bottom option for the tab component.

dvoiss:
I wanted tabs on the bottom so I've made an addition to support that. 

I'm new to React and have been away from front-end web dev for a while so if anything isn't standard or best practice let me know what needs to be changed. Thanks!

This is an image of `http://react-toolbox.com/#/components/tabs` with `tabsOnBottom` set to `true`.

```javascript
<Tabs tabsOnBottom={true} index={this.state.index} onChange={this.handleTabChange}>
```

screen shot 2016-05-04 at 2 06 09 am

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -8,6 +8,7 @@ class Tabs extends React.Component {
     children: React.PropTypes.node,
     className: React.PropTypes.string,
     disableAnimatedBottomBorder: React.PropTypes.bool,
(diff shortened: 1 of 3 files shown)


javivelasco:
Unfortunately I think this is out of the material design spec isn't it? https://www.google.com/design/spec/components/tabs.html#tabs-specs I think it shouldn't be merged officially, sorry :(

Issue #504, opened 2016-05-05
Can't install npm module from GitHub


jeffreywescott:
Thanks again for this great library. :-)

I was hoping to install react-toolbox via npm using `github:react-toolbox/react-toolbox` (because there have been some bug fixes that have been merged but not yet released that are pretty critical). Unfortunately, because you don't commit your `lib` folder, it doesn't work.

It's pretty common practice for people who want to ride the "bleeding edge" to install npm modules from GitHub. However, because you have `npm run build` as a `prepublish`, it won't work. Any chance you could follow the `npm` recommendations here and commit your `lib` folder?

lzrski:
Shameless plug: try https://github.com/lzrski/npm-git-install and let me know if it works for you :)

PR #505, opened 2016-05-05
allow react-toolbox to be npm-install'ed from GitHub

jeffreywescott:
Addresses #504.


felipeleusin:
Any love for this @javivelasco ? Would really like to be able to use this on some of my projects

javivelasco:
Yeah, but since the next release will be done over the new style approach I don't know if the PR should be done here or we should just wait until the branch is ready

javivelasco:
Coming in the next release!

PR #509, opened 2016-05-10
Media queries for dialogs to set wider widths for smaller screens

johnnynotsolucky:
Dialogs are squashed on smaller devices. Therefore, I've added media queries to address this: `large` is unchanged; `normal` becomes `96vw` on `$layout-breakpoint-xs`; And, `small` becomes `75vw` on `$layout-breakpoint-xs` and `50vw` on `$layout-breakpoint-sm-tablet`

### DIFF
--- a/components/dialog/style.scss
+++ b/components/dialog/style.scss
@@ -23,10 +23,22 @@
 
 .small {
   width: 30vw;
+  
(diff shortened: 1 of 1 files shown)


javivelasco:
Great, thanks!

PR #510, opened 2016-05-10
removed the redundant import of mixins.scss and updated the dependencies

gbhasha:
mixins.scss is already imported in base.scss, hence removing  the redundant import.


javivelasco:
Thanks!

Issue #520, opened 2016-05-19
Input label not staying on top when containing input after losing focus...


steviesama:
I was gonna use the multiline input to write in some text input, but when it loses focus, the label which has animated to the top, returns to input area of the element.

Here it is animated up top: 
![working](http://i.imgur.com/FIclFZ0.png)

And here it stacks on top of the user input:
![not-working](http://i.imgur.com/sI83eWm.png)

Is there a way to remove this behavior, or is this not intended?

**EDIT:**
If I set a value as a prop, it stays on top...I noticed in the playground app, the label doesn't have any problems. Doesn't make much sense unless there is extra steps that have to be taken to ensure this behavior.

dfischer:
You need to set state/value

steviesama:
State isn't the problem. The label doesn't need to change. It's the the animated behavior. I added state just to test it anyway, but still nothing. Shown in the shot below.

![tested](http://i.imgur.com/Fr1kkDW.png)

The label still dropped back down and overlaps the input if you remove focus from the input box.

Focus as you can see via the text cursor in the input:
![focus](http://i.imgur.com/Dl2HCju.png)

And overlapping when it loses focus:
![nofocus](http://i.imgur.com/wIQJFsr.png)

KerenChandran:
This is a repeated issue, #465. PR #523 somewhat fixes this issue, but is waiting approval.

As mentioned above, you did not set the value parameter of the input field. Furthermore, the value was supposed to have the state, not the label. Treat the state as something that you can mutate, in this case the user's text in the input field. The label is usually static so it can be hard coded string.

To fix this code, include the following:

```js
constructor() { //Don't think you need to include props, unless you're using it in the constructor.
  this.state = {time: 0};
  this.onTimeChange = this.onTimeChange.bind(this); //Required since events are not autobinded in ES6 react
}

//You can make this generic to handle more input cases, but I'll leave that up to you
onTimeChange(value) {
  this.setState({time: value});
}

render() {
  //Change your input line to be the following
  <Input label="Start Timer" onTimeChange={this.onTimeChange} value={this.state.time} />
}
```

KerenChandran:
Furthermore, #465 had issues cause they wanted to use refs. Here it seems more like an implementation issue, unless you were planning on using refs to access input control.

javivelasco:
[...]

PR #523, opened 2016-05-21
Added forceFloatLabel property to input

### LINKED ISSUE #465: Input component requires value property if you use a label
When I want to use the Input component with a label and I don't set the value using state, the label will not stay in its up position when the element is not in focus anymore. I prefer to use refs instead of state to get the value out of my inputs, but this option would not work properly.

In order to replicate the problem:
- Make an Input with a label, but without a value property
- Try typing in the input, this pushes the label out of the way, as expected
- Now when the input loses focus, the label returns to its original position, in the same place that the input value is

If it makes any difference I am using react-toolbox with Meteor and am using version 0.16.2.

KerenChandran:
Fixes #465 

Added forceFloatLabel as mentioned in issue to force label to float if user wishes to use null. Default value is set to false and does not interfere with current behaviour.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -10,6 +10,7 @@ class Input extends React.Component {
     disabled: React.PropTypes.bool,
     error: React.PropTypes.string,
     floating: React.PropTypes.bool,
+    forceFloatLabel: React.PropTypes.bool,
     hint: React.PropTypes.string,
     icon: React.PropTypes.oneOfType([
       React.PropTypes.string,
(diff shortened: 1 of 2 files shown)


KerenChandran:
Build failed due to Redbox-react (1.2.5 installed since package.json had ^1.2.4) peer dependency requirement (react ^15.1.0). Updated it and didn't see any odd behaviours.

javivelasco:
Please, see #520!

PR #524, opened 2016-05-23
Fix display issue with slider

Valentin-Seehausen:
The innerprogress seemed moved on certain browser versions.
Added `left: 0` property to .innerprogress


Valentin-Seehausen:
Further investigation: The problem occurs, when slider is in an `text-align: center` container.

javivelasco:
May you give a use case where this is not working??

javivelasco:
I'll reopen then to reduce noise. Also try to check against the new release!

PR #526, opened 2016-05-26
Media-queries for the max-height of the app-bar on mobile devices

dannydorfel:
As described in https://www.google.com/design/spec/layout/structure.html#structure-app-bar


javivelasco:
Thanks! :)

PR #537, opened 2016-06-02
Ie slider fix

JohnBicknell:
Interacting with the Slider was sending NaN to the change handler for me in IE 10 and 11 which doesn't seem to have window.scrollX / Y. Change uses an alternative IE property and fixed it for me.


javivelasco:
Thanks!

PR #538, opened 2016-06-03
small fix for autocomplete component

yuristrelets:
For `multiple=false` components value is broken, because it isn't extracted from Map.

If I'm trying to use key/value object as source I get key instead value (in input) as expected. But after full render of component value is visible again.


balkonsmoke:
awesome pr!

PR #546, opened 2016-06-06
feat(datepicker,timepicker): add support for `name` prop

mgcrea:
Just added `name`-prop support for both Input, Timepicker and Datepicker components.

This enables to use `ev.target.name` to easily get the control name of the control form element. Thus being able to use the same `onChange` handler for several inputs/components.

This makes Datepicker/Timepicker act a little bit more like regular inputs (which is what they are ultimately, before being heavily decorated).

More info on MDN:

`name: The name of the control, which is submitted with the form data.`

- https://developer.mozilla.org/en/docs/Web/HTML/Element/Input
- https://developer.mozilla.org/en/docs/Web/HTML/Element/Button

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -23,6 +23,7 @@ const factory = (Input, DatePickerDialog) => {
       ]),
       inputClassName: PropTypes.string,
       inputFormat: PropTypes.func,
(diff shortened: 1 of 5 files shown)


javivelasco:
Thanks!

PR #548, opened 2016-06-07
Update TypeScript Definitions and fix npm typings

hsrobflavorus:
I went through and split out the react-toolbox.d.ts TypeScript definition file, which had all the modules declared in an ambient context (not the right way to do it), into separate external module files for each component. Now the TS definitions should actually work right out of the box when react-toolbox is installed via npm.


javivelasco:
Thanks! :D

PR #551, opened 2016-06-08
Snackbar has `transition: all` and it seems un-needed...

Klathmon:
The `transition: all` on the `<Snackbar>` element was causing weird things like my color to transition from one color to another when it changed.

I felt that this was a bit un-intuitive. I expected the element to only apply the transition timings to the transforms it does to "pop up" and "close down".

This is a very small change to have it act that way.

Feel free to close this PR if you feel it's working as intended as is, and if you need me to i can run the build command and re-commit the change with the `lib` dir built as well.


javivelasco:
Looks nice, thanks! 💃

Issue #178, opened 2015-11-30
Can't interact with page while Snackbar is visible


alltom:
While a Snackbar is visible, the rest of the page doesn't get mouse events.

I was going to add a persistent bar reminding the user to reload the page if I detect that there's a newer version of the site, but that prevented people from using the site until they dismissed the bar.

javivelasco:
I think I know why is happening. I'm going to fix it tonight, thanks for reporting @alltom !!

javivelasco:
Not yet, I've been quite busy last week. I'll try to move it later today!

wrench7:
Thanks for adding that fix :-)

Worth noting that the pointer-events prop that you added won't work in IE < 11, which made me wonder what baseline browser config react-toolbox supports?

javivelasco:
Truth is I'm not worrying too much about cross-browser issues. I'd like to support lower versions than IE11 in a new iteration soon. For now I can guarantee react-toolbox works in all latest Chrome, Safari and Firefox since they are the ones I can test physically.

PR #552, opened 2016-06-08
Fix pointer-events for Overlay in 1.0

Klathmon:
Looking at git-blame it looks like in commit [#f6d2d3f9](https://github.com/react-toolbox/react-toolbox/commit/f6d2d3f95d2c2771b4ebd7fe2e7f992a46dd6a39) broke the handling of the `transparent` prop and `pointer-events` in the `<Overlay>` element. (basically re-opening #178)

The sass had `&.invisible > *:not(.overlay) {` which looks like it should have been `&.invisible > *:not(.backdrop) {` with the updated sass for the theme stuff.

This fixes the issue for me. Again let me know if you want me to run the build on my own, i can do that before this is merged. (and sorry about the 3 commits, one from the other PR got in there and I had to revert it)

### DIFF
--- a/components/overlay/theme.scss
+++ b/components/overlay/theme.scss
@@ -16,7 +16,7 @@
   align-items: center;
   justify-content: center;
   pointer-events: none;
-  &.invisible > *:not(.overlay) {
+  &.invisible > *:not(.backdrop) {
     pointer-events: all;
   }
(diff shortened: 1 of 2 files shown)


javivelasco:
Thanks! 👯

PR #554, opened 2016-06-09
Patch feat name prop

mgcrea:
Added name prop support for the `Dropdown` component. Since the emitted target is currently a `<li>` and not an input, I could not leverage the input `name` attribute.

Some minor propTypes re-ordering following #546 to fix eslint warnings.


javivelasco:
Makes sense, thanks!

PR #555, opened 2016-06-09
[Fixes #550] Add PropType checking to Menu's position and update docs

### LINKED ISSUE #550: IconMenu positioning not working since 1.0.0?
Not sure if I've missed something but the `IconMenu` demo (http://react-toolbox.com/#/components/menu) seems broken, the menu does impact the flow of DOM elements when closed (eg. hidden but still there in the DOM), shouldn't the position be absolute-like?

Also tried to play with the `position` attr/prop, and nothing changes.

Is it a regression or has it always behaved like this? Is this open for a PR?

Danno040:
I got burned by Issue #550 when upgrading to 1.0.0. The position property on Menu got moved from dashes to camelCase, but the docs weren't updated.

This PR updates the docs, and adds a PropType check on position.

### DIFF
--- a/components/menu/Menu.js
+++ b/components/menu/Menu.js
@@ -25,7 +25,7 @@ const factory = (MenuItem) => {
       onSelect: PropTypes.func,
       onShow: PropTypes.func,
       outline: PropTypes.bool,
-      position: PropTypes.string,
+      position: PropTypes.oneOf(Object.keys(POSITION).map(key => POSITION[key])),
       ripple: PropTypes.bool,
       selectable: PropTypes.bool,
       selected: PropTypes.any,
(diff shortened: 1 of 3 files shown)


components/menu/Menu.js
```diff
@@ -25,7 +25,10 @@ const factory = (MenuItem) => {
       onSelect: PropTypes.func,
       onShow: PropTypes.func,
       outline: PropTypes.bool,
-      position: PropTypes.string,
+      position: PropTypes.oneOf([
```
jingyuyao: Wouldn't it be better to do something like this?
```
position: PropTypes.oneOf(Object.keys(POSITION).map(key => POSITION[key])),
```

Danno040: I wasn't sure the support for Object.keys, so I hesitated to use it. Its ES5, so I updated the code to use your snippet.

javivelasco:
Wow, sorry about that!

PR #556, opened 2016-06-10
Fix MenuItem onclick called without event

paulocmoreno:
In all other cases the event listener receives the event. I need to receive the event to stop its propagation.


javivelasco:
Great work!

PR #562, opened 2016-06-13
Fix hanging timeout on <Snackbar/> element

Klathmon:
The `<Snackbar/>` element sets a timeout if you set the timeout and onTimeout props on the element. If the element is unmounted before those props fire, they can possibly fire on an un-mounted parent element.

This fix just puts an componentWillUnmount handler on the Snackbar element and has it clear it's timeout when it unmounts which prevents that issue.


javivelasco:
Great! Totally missed it!

Issue #533, opened 2016-05-31
Calendar 'next month' button skipping a month


ro-savage:
There appears to be a bug with the 'next' and 'previous' buttons on the calendar date picker.

If the current day of the month, is larger that the total amount of days in the next month, the 'next month' function skips a month.

> May 31 -> July (Skips June which has 30 days)
> June 30 -> July (July has 31 days)
> July 31 -> August (August has 31 days)
> August 31 -> October (Skips September which has 31 days)
> September 30 -> October (October has 31 days)
> October 31 -> December (Skips November which has 30 days)

If you do the reverse, e.g. 31 December to backward. It will not work on first press, on second press will take you to the correct month.

Video: http://recordit.co/CGZnQSIC7r

Can take a look and submit a PR if needed.

Is it better to make a PR to current master? Or wait until this big theme change occurs?

ro-savage:
PR #568 fixes this issue for v1.0.0

PR #568, opened 2016-06-15
Fixed bug with changes months on datepicker

ro-savage:
This is an issue with using the `.setMonth` method when the current day value is larger than the largest day value of the month being set.

When setting a new month, we now set the date to the first day of the month.

E.g.
Adding 1 month to `Jan 31`, means it tries to set the date to `Feb 31` which doesn't exist and therefore it instead sets the date to `Mar 02`.

```
var theDate = new Date('01 31 2016')
theDate.setMonth(1)
console.log(theDate) // 03 02 2016

var theNewDate = new Date('01 31 2016')
theNewDate.setMonth(1, 1)
console.log(theNewDate) // 02 01 2016
```

Fixes issue #533

### DIFF
--- a/components/utils/time.js
+++ b/components/utils/time.js
@@ -104,7 +104,7 @@ const time = {
 
   addMonths (d, months) {
     const newDate = this.clone(d);
-    newDate.setMonth(d.getMonth() + months);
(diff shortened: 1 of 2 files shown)


javivelasco:
Bien visto! 😄

PR #569, opened 2016-06-15
Feature - Autocomplete - Added autoselect first option on enter

ro-savage:
This modifies the autocomplete so that when you press `enter` it will automatically select the top option.

This is overwritten if the user hovers over an option and press `enter`.

Not sure if this is the best way to do this. Happy to hear feedback.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -112,8 +112,17 @@ const factory = (Chip, Input) => {
    };
 
    handleQueryKeyUp = (event) => {
-     if (event.which === 13 && this.state.active) this.select(this.state.active, event);
+     if (event.which === 13) {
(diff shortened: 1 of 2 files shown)


javivelasco:
I need to take a closer look, gonna check during the weekend :)

ro-savage:
This could certainly be implemented better. I tried a couple things, but kept running into issues with setting state and re-renders. 

So decided that it was a fair assumption that if the user pressed 'enter' they wanted to just select whatever was on top (or had been selected with the keyboard).

Having the 'active' state set to whatever is currently at the top would make logical sense but was causing multi re-renders.

I'm been trying to make our forms + react-toolbox behave better for keyboard only use. Next thing is the date picker and being able to type in a date.

javivelasco:
I think it's good enough! 👍

PR #571, opened 2016-06-15
Fix: IE11 displaying calendar incorrectly

ro-savage:
The calendar is out of line for IE11 due to some extra padding. Removing the padding didn't seem to have an effect in Chrome and Firefox (did not test mobile browsers) and fixed the issue on IE11.

Have add topbottom and leftright padding. Only topbottom is used by default, but leftright can be overwritten.

Before:
broken

After:
fix

(Other styling is different as fixed image was taken off my project, and the broken image is from react-toolbox docs)


javivelasco:
Nice!

PR #572, opened 2016-06-15
Add 'suggestionMatch' prop to Autocomplete component

### LINKED ISSUE #566: Autocomplete suggestions should use .includes()
Using `.startsWith()` on this line:

https://github.com/react-toolbox/react-toolbox/blob/afde8b1aa21122aabac1145346ad74f2c4a1782e/components/autocomplete/Autocomplete.js#L154

...is of extremely limited usefulness because it will only suggest matches that start with the query. 

Can this just be switched to use `.includes()` instead (to match the query anywhere in the string)? 

Would you want a prop added to the Autocomplete component to control this behavior? 

Am willing to do a PR if needed, just wanted to see if this has been discussed before and which approach would be favored.

thearrow:
Fixes #566 

Add `suggestionMatch` prop to Autocomplete component to determine how suggestions are matched.

It can be:

- `start`: query matches the start of a suggestion - eg. `ne` will match `new zealand`

- `anywhere`: query matches anywhere inside the suggestion - eg. `land` will match `new zealand` and `thailand` and `england`

- `word`: query matches the start of a word in the suggestion - eg. `zea` will match `new zealand`

Defaults to `start` to maintain consistency with current behavior.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -26,6 +26,7 @@ const factory = (Chip, Input) => {
      selectedPosition: PropTypes.oneOf(['above', 'below']),
      showSuggestionsWhenValueIsSet: PropTypes.bool,
      source: PropTypes.any,
+     suggestionMatch: PropTypes.oneOf(['start', 'anywhere', 'word']),
      theme: PropTypes.shape({
        active: PropTypes.string,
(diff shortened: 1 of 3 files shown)


ro-savage:
Awesome work!

javivelasco:
❤️

PR #573, opened 2016-06-16
Added ability to export the DatePickerDialog separately from the DatePicker

ro-savage:
Our project required the calendar (`DatePickerDialog`) without the entire `DatePicker` that went with it.

With the change to v1.0.0 this was no longer being exported, so this update simply exports `DatePickerDialog` and `DatePicker`


javivelasco:
👍 !!

PR #579, opened 2016-06-16
Fix #575: Use classname 'material-icons' for FontIcon if children is a string

### LINKED ISSUE #575: FontIcon with children is missing classname 'material-icons'
Hi,

The example in http://react-toolbox.com/#/components/font_icon is not working. The last item shows the text 'star' instead of the icon.

Viewing the code, I can see this line:

```
className={classnames({'material-icons': typeof value === 'string'}, className)}
```

Should it also have classname 'material-icons' when children are a string?

Otherwise, I must use:
```
<FontIcon className='material-icons'>star</FontIcon>
```

BaoDelta:
It fixes #575.


javivelasco:
👏

PR #582, opened 2016-06-20
Bugfix - DatePicker and TimePicker pass unknown props and allow enter key

ro-savage:
Updates DatePicker and TimePicker to have the same behaviour as other components with an Input and pass unknown properties to the Input (e.g. `tabIndex`).

It also adds a KeyPress event to the input, that opens the dialog on `enter` and passes the event to any KeyPress passed as a prop.

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -28,6 +28,7 @@ const factory = (Input, DatePickerDialog) => {
       minDate: PropTypes.object,
       name: PropTypes.string,
       onChange: PropTypes.func,
+      onKeyPress: PropTypes.func,
       theme: PropTypes.shape({
         input: PropTypes.string
       }),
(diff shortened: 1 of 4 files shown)


components/date_picker/DatePicker.js
```diff
@@ -67,12 +76,14 @@ const factory = (Input, DatePickerDialog) => {
             className={classnames(this.props.theme.input, {[inputClassName]: inputClassName })}
             error={this.props.error}
             onMouseDown={this.handleInputMouseDown}
+            onKeyPress={this.handleInputKeyPress}
```
javivelasco: It this actually working? If you spread `props` in the last line and provide `onKeyPress` is provided, I think it's going to override that `handleInputKeyPress` assignment. You should either extract that prop at line 68 or spread the `others` variables as the first prop passed down.

ro-savage: Good pickup. Spreading props at the top now.

javivelasco:
Thanks! 👍 !

PR #588, opened 2016-06-23
chore(dropdown): set value to empty string to prevent uncontrolled er…

mgcrea:
With the current implementation, `react@^15` spills warnings about uncontrolled inputs:

```
warning.js?746c:44 Warning: Input is changing a controlled input of type null to be uncontrolled. Input elements should not switch from controlled to uncontrolled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

Setting the value to be an empty string `''` instead of `false` does fix that.


javivelasco:
Thanks!

Issue #278, opened 2016-01-24
Multiline input doesn't resize according to the number of lines


SimonSelg:
Hey,

according to the docs, the multiline inputs textarea will
> grows and shrink according to the number of lines.

I assume this means something like [this](http://www.jacklmoore.com/autosize/).

However, this isn't implemented in the current version of react-toolbox. Is this just an mistake in the docs, or is it just not implemented yet? If that's the case, I'd create an PR for it.

Cheers,
Simon

mrogelja:
+1 for a PR

javivelasco:
Not implemented yet but it would be very nice!

madisvain:
this would be a nice feature to add

zhaoyao91:
+1

PR #590, opened 2016-06-24
Add autoresize to multiline inputs

SimonSelg:
This implements auto resizing for the multiline inputs, as discussed in #278.


javivelasco:
It's merged! I did it via command line to change the if/else for a ternary. 
Great work! thanks @SimonSelg :)

PR #599, opened 2016-06-26
Better support in typescript

ixrock:
- added missed exports declarations (`Layout`, `Ripple`)
- export components by name besides default export (as provided in source files)
- fixed missed references in navigation component for `ButtonProps` and `LinkProps`
- removed duplicating declarations in `components/`


JeroenNelen:
@ixrock I suppose this is backwards compatible?
If so then maybe this can be merged @javivelasco.
The PRQ only affects definition files. 
Then afterwards I will probably also make a typescript-related PRQ, the onChange definition for the input is missing the event parameter.

Kind regards,

PR #601, opened 2016-06-27
Bugfix: Reset overflow when unmounting an overlay

strindhaug:
Fixes issue where scrolling is disabled after overlay is unmounted.

When used in a menu componentWillUpdate may not be called when clicking on a menu item, resetting overflow on body on componentWillUnmount ensures you can scroll after using the menu.


javivelasco:
Great, thanks!!

PR #602, opened 2016-06-27
missing MENU for some reason in identifiers

lwhorton:
Not quite sure why this went missing - perhaps a bad merge?


aft-luke:
@javivelasco any word on getting this in? I would hate to have to fork and set that all up if I can avoid it :/

javivelasco:
Yeah, merging and releasing asap. I don't know how that had happened :/

PR #617, opened 2016-07-07
Fix Number.isNaN TypeError on IE

InfernoZeus:
On IE11, some of the components in the docs demo page don't work - Autocomplete and Slider are two examples. They fail with the following message:
> TypeError: Object doesn't support property or method 'isNaN'

I've fixed these by replacing Number.isNaN with isNaN, which is much more widely supported. Of course, it behaves differently when passed values that aren't of type number. I'm not quite sure what ```valuePresent``` is used for, but I hope the switch is acceptable. If not, an alternative Number.isNaN can be added quite easily.

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -112,7 +112,7 @@ const factory = (FontIcon) => {
         [theme.withIcon]: icon
       }, this.props.className);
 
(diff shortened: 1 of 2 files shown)


javivelasco:
I think this should work without any bad side effects. Thanks!

PR #631, opened 2016-07-11
Document the label theme for the input component

randycoulman:
The `input` component has a `label` theme setting, but it wasn’t documented.


javivelasco:
Thanks!

PR #635, opened 2016-07-12
added default callbacks for onEscKeyDown & onOverlayClick methods

rajdee:
I added default callbacks for onEscKeyDown & onOverlayClick methods because we can't set state from parent component


javivelasco:
I see the problem, but IMO it would be better to go with other approach. For example, the same as `Dialog` use. Maybe the state should be extracted to a property or, in case we want to keep state, we should accept an `active` property to sync internal state with it. This way it would be optional.

javivelasco:
Closed in favor of #693

PR #644, opened 2016-07-14
Update README.md

JamesAlias:
In section *Roboto Font and Material Design Icons* update link to Roboto font.
Now links to `https://fonts.google.com/specimen/Roboto`.


javivelasco:
Thanks!

PR #646, opened 2016-07-15
Fix autocomplete activating on react>=15.1 if no direction

farwayer:
`Autocomplete` will not show suggestions while input is first time activated if no `direction` specified. Bug is here https://github.com/react-toolbox/react-toolbox/blob/dev/components/autocomplete/Autocomplete.js#L71

All worked ok on `react<15.1` due the bug was in react core (twice update, https://github.com/facebook/react/pull/6650). On `react>=15.1' `Autocomplete` is not activated correctly.


javivelasco:
I've realized and fixed while working in #666 but since it needs to wait a bit to be merged, I'm accepting this and releasing as a patch. Thanks!

PR #650, opened 2016-07-17
Typescript typings

normano64:
Redid the typings for the project as I guess they were accidentally removed with a `npm run clean` and not up to date. If you have any input or problems with it, do tell.

- [x] Add all components and their props
- [x] Add theme interfaces for the components that support themes.
- [x] Add JSDoc comments to all props.

As it is right now, you'll get a `TS2656` error if you import like `import Avatar from "react-toolbox"` instead of `import Avatar from "react-toolbox/lib/avatar"`. Don't know why or how to fix it other than moving the typings to `dt`.


index.d.ts
```diff
@@ -1,57 +1,1324 @@
...
+    link?: string;
+  }
+  interface LinkProps extends Props {
+    active?: boolean;
+    children?: React.ReactNode;
+    count?: number;
+    icon?: React.ReactNode | string;
+    label?: string;
+    theme?: LinkThemeProps;
+  }
```
giacomorebonato: Not sure, but I think that the property 'href' which is implicit in the codebase, should be present here.

javivelasco:
Great!

PR #658, opened 2016-07-21
Only add Input auto-resize handler to multiline inputs

InfernoZeus:
Without this, resizing a window which contains an `input type=file` causes the Browse button to be resized. I don't think there's any reason to monitor resizing for single line inputs?


javivelasco:
Thanks!

PR #659, opened 2016-07-21
402 max date not taken into account

### LINKED ISSUE #402: DatePicker maxDate in past not taken into account when dialog is shown
If I set a `maxDate` that is in the past (e.g., 1995-03-21), when I click on the DatePicker field:

- the date still defaults to _today_, even though that date is not allowed
- it's still possible to move past the `maxDate`
  - the year drop-down still shows years after 1995
  - the year 1995 still shows months 04-12

This forces the user to navigate backwards year-by-year and / or month-by-month, which takes quite a long time.

_At a minimum_, the dialog should open somewhere _before_ `maxDate` (perhaps at the midpoint between `minDate` and `maxDate`?).

While I haven't tested it, I suspect something similar is true for `minDate`.

AlexGilleran:
Fixes #402.

I've just made the date set on mount / `propsWillChange` go through what was called `handleCalendarChange`, although now it handles all changes I've renamed it `handleNewDate`. This means that if you set a max/min date, the first time you open the date picker the initial selected date will be within that range. I've implemented a simple check to sanitise an out-of-bounds date to the closest out of `minDate` or `maxDate`.

I've also written some tests but it required me exporting `DatePickerDialog` the same way that you've exported non-themed components for testing purposes, hope that's ok.

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -115,5 +115,8 @@ const DatePickerDialog = datePickerDialogFactory(InjectDialog, Calendar);
 const DatePicker = factory(InjectInput, DatePickerDialog);
 
 export default themr(DATE_PICKER)(DatePicker);
-export { factory as datePickerFactory };
+export {
+  DatePickerDialog as DatePickerDialog,
+  factory as datePickerFactory
+};
 export { DatePicker };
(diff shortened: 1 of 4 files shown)


javivelasco:
This is amazing! It's great that you wrote some tests, thank you 💃 
Merged, I've just needed to fix some linting errors.

AlexGilleran:
Happy to help :)

PR #661, opened 2016-07-21
Ensure theme gets passed down to TimePicker child components

### LINKED ISSUE #660: TimePicker theme not getting applied to children
It seems like the `TimePicker` component is not correctly passing down its `theme` to its children.

It looks like the destructuring of `theme` in [line 62 of `TimePicker`](https://github.com/react-toolbox/react-toolbox/blob/dev/components/time_picker/TimePicker.js#L62) means it does not get splatted down in [line 67](https://github.com/react-toolbox/react-toolbox/blob/dev/components/time_picker/TimePicker.js#L67).

If you compare with `DatePicker`, it does *not* get destructured in [line 71](https://github.com/react-toolbox/react-toolbox/blob/dev/components/date_picker/DatePicker.js#L71) so it *does* get splatted in [line 79](https://github.com/react-toolbox/react-toolbox/blob/dev/components/date_picker/DatePicker.js#L79).

conroywhitney:
Fixes #660 by ensuring that the `theme` prop gets passed down to child components.

Follows the same pattern that `DatePicker` does by *not* destructuring `theme` from `this.props` but instead referring to it as `this.props.theme` when setting the `className` on the `<Input />` component.

### DIFF
--- a/components/time_picker/TimePicker.js
+++ b/components/time_picker/TimePicker.js
@@ -59,13 +59,13 @@ const factory = (TimePickerDialog, Input) => {
     };
 
     render () {
-      const { value, format, inputClassName, onEscKeyDown, onOverlayClick, theme, ...others } = this.props;
+      const { value, format, inputClassName, onEscKeyDown, onOverlayClick, ...others } = this.props;
       const formattedTime = value ? time.formatTime(value, format) : '';
       return (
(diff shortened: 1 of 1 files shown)


javivelasco:
Right! Thanks :) !

Issue #634, opened 2016-07-12
How do I override button font size


shineli1984:
I was trying to change the overall sizes by changing $unit, but I noticed that raised button has fixed `font-size: 1.4rem` which is hardcoded in typo-button mixin in the global mixin.scss

Is there a way to override this?

kossel:
you may want to create a scss selector which match the element
```
<Button className={styles.parent} />
```

```
.parent{
    button {
      font-size: $larger-font;  //or whatever the new size
  }
}
```

javivelasco:
PR is merged and it makes sense!

PR #664, opened 2016-07-22
Made general mixins properties depend on $unit value

xmityaz:
I've faced the similar issue as #634.
So I thought it can be useful to have a possibility to overwrite _mixins absolute values with the $unit property. Sure if only it wasn't made for a reason.


javivelasco:
Yeah, it makes a lot of sense. I'm going to merge it now to be released as a patch, but for the migration I'm currently working on #666 mixins would mostly disappear in favor of custom properties. Thanks!

Issue #647, opened 2016-07-15
Table select issue


farwayer:
Select logic in `Table` was broken by this commit https://github.com/react-toolbox/react-toolbox/commit/33ff49119b49227d269b28cbc2caf1d57c31939e#diff-524a577bb5ecc86904df93f328ecf9e6L40

PR #667, opened 2016-07-24
Fix multiselectable and unselect rows issues on Table component

vmsoftware:
Currently, multiselectable and unselect features don't work properly on Table component. You can see it at [http://react-toolbox.com/#/components/table](http://react-toolbox.com/#/components/table).

The problem was in `Table.js` file with the use of `splice` function because `splice` return a new array with the item that is removed from the original array. Then always return the last item selected instead of the actual items plus the new selected item (when multiselectable is enabled, of course).

I changed the file to use `filter` to solve the problem and restructure the if conditions.

This PR solve #647

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ build
 node_modules
 npm-debug.log
 .idea
+.DS_Store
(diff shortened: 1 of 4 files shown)


javivelasco:
I've changed a bit the code to pass the linter and make it more readable, but essentially it's doing the same as your fix. Thanks! :)

Issue #668, opened 2016-07-24
Theming nested components (<ListItem />'s children)


slindberg:
New to react-toolbox, thanks for all the hard work! I'm finding it fairly straightforward to compose components, but I'm having trouble with theming nested components, e.g. `<ListItem />` → `<ListItemLayout />` → `<ListItemContent />` → `<ListItemText />`.

My understanding is that `react-css-themr` provides a way to define _default_ styles for components, but does not address overriding styles for particular components. For this, we must pass the `theme` prop to RT components with a class definition that matches the documented root class, e.g. `.item { ... }`.

This works great one level deep, but I do not understand how to style nested RT components. For example, how do I change the text color of a `ListItemText` component, when all I'm rendering in my component is `<ListItem theme={theme} ... />`? Of course, I could just select the correct elements by name (`.item > span > span { color: red; }`), but this seems to defeat the point of CSS modules, and is brittle.

From [the docs](https://github.com/react-toolbox/react-toolbox/tree/dev/components/list#theme-1), it looks like you should be able to provide top level class definitions for nested components in the theme of the the component you are rendering, i.e. defining a `.itemText { ... }` block in the theme passed to `<ListItem />`. However, in this case the theme doesn't appear to be passed through to the `<ListItemText />` component. How are these theme names/classes intended to be used? Thanks!

slindberg:
Is this perhaps specific to list items? I could submit a PR that passes the `this.props.theme` through on [this line](https://github.com/react-toolbox/react-toolbox/blob/dev/components/list/ListItemLayout.js#L26) and [this line](https://github.com/react-toolbox/react-toolbox/blob/dev/components/list/ListItemContent.js#L40) (and others). It seems like this would have the desired effect, but I want to make sure this is the intent and I'm not missing something.

javivelasco:
Hi @slindberg! You got it right from what I read :) In this approach subcomponents are intended to use the same theme object than the parent. A caveat is that they can't share the same classNames and that's why some of them are namespaced. The theme is not directly passed down because by default it can be passed in the export definition. So the overrides can be provided straight when you use the subcomponents.

If a component catches and render the subcomponent if should spread the props. If not, it should be considered a bug.

slindberg:
Great, submitted #669 to address the issue.

PR #669, opened 2016-07-25
Pass theme down to list item child components for correct classnames

slindberg:
See #668 

This allows theming the `<ListItemContent />`, `<ListItemText />`, and `<ListItemAction />` components from the theme passed into the parent `<ListItem />`, using the `.itemContextRoot`, `.itemText`, and `.itemAction` classes.


javivelasco:
Great!

PR #670, opened 2016-07-26
Promote ripple component to its own composite layer only when it is active

avdeveloper:
Currently, IconButtons are being rendered on their own composite layer because RippleComponent is in the render tree even though it's invisible. This is a problem because if you have several hundreds of instances, scrolling becomes janky. Therefore, we should exclude it from the render tree via `display: none` and only show it when the user clicks on the button.

**Before:**
screen shot 2016-07-26 at 4 58 31 pm

**After:**
screen shot 2016-07-26 at 4 58 48 pm

### DIFF
--- a/components/ripple/Ripple.js
+++ b/components/ripple/Ripple.js
@@ -73,6 +73,9 @@ const rippleFactory = (options = {}) => {
       handleEnd = () => {
         document.removeEventListener(this.touch ? 'touchend' : 'mouseup', this.handleEnd);
         this.setState({active: false});
+        setTimeout(nil => {
+            this.setState({ visible: false });
+        }, 1000);
       };
 
       start = ({pageX, pageY}, touch = false) => {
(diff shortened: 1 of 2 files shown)


javivelasco:
@avdeveloper thank you so much for noticing and for your PR. I think there are some problems with the approach your proposed IMO:

- Setting a timeout to make invisible the ripple feels kind of hacky. If somebody wants to change the animation duration to say 2 seconds, it wouldn't work anymore. We can take advantage of the animation event listeners to get noticed when the animation actually ended and the ripple can be hidden.
- I would avoid the `!important` flag when possible. Although this solves the composite layer problem, the useless node is still there.

Based on your PR I've just made [this commit](https://github.com/react-toolbox/react-toolbox/commit/032840e74a7a737596914cd67a2946284beca859). I'm doing basically the same: adding a new state property to know when the ripple is visible. The main difference is that if the component is not visible it remains unmounted and to know when it should be removed I'm using the animation event handlers.

I've tested and it work :) Sorry this is not getting directly merged but the credit is all yours.
Thanks!!! 🚀

PR #677, opened 2016-07-29
pass theme from layout to children

bschnelle:
similar to #669.


javivelasco:
Makes sense, thanks! 💃

Issue #623, opened 2016-07-09
localisation / i18n support


joeblew99:
These components are very high quality and very reusable, but i frankly need the ability to do i18n on them to use this in real world projects.  
So, i am asking for some support on this.

Generally you need to project a way to do:
- text translation
- datetime, currency and number formatting
- RTL and LTR layout even

Its a difficult task and big, but others have done it for React.
SO far i have found this, and it looks well supported.
https://github.com/yahoo/react-intl
- tackles the hard aspects
- lots of committers
- looks good to me so far
- considers the workflow. This is very important
- backed by format.js (http://formatjs.io/integrations/)

But how to integrate it with react toolbox in a sustainable way is really important.
There seems to be utter confusion about how best do this. Use Props, Context or Mixins.

I would very much appreciate feedback on this.

Also i am curious if others feel this is an important thing to them ? 

There has been a hack done in the past. Its does not really do datetime formatting 
https://github.com/react-toolbox/react-toolbox/issues/229

javivelasco:
I think there are components like the `DatePicker` where localisation should be definitively taken into account. #680 does a great work and it's being published today with a new release. I think the best we can do to avoid messing up too much with this is to move any hardcoded message to a property. For example, dialogs for pickers should have the message coming from properties. We should identify all components where this happens and extract messages to props.

For other components like the DatePicker it's more complicated but it's done anyway.

javivelasco:
I think this can be closed. Every component should be flexible enough to allow to pass custom translated text and the most problematic (which was `DatePicker`) is already localised. If you find a component that cannot be changed please open a new issue with the details about the specific case.

PR #680, opened 2016-08-01
Added support for locale on DatePicker dialog component

aabilio:
Updated components, docs and spec.


aabilio:
This is not a PR to address #623, just a simple way to provide strings to DatePicker to show months and days in other languages (with some predefined languages loaded).

javivelasco:
Beautiful work! I think this is very useful and does not create any breaking changes so happy to merge! Thanks 💃 !

sanfilippopablo:
This is great. When is this going to be on NPM?

aabilio:
I just realize that action buttons ('Cancel' and 'OK') need to be translated too. I was thinking on inserting it in locale prop too (pro: easy api for Date Picker component) but I think that inserting new properties like okButtonLabel, cancelButtonLabel, dismissButtonLabel... could make sense too (pro: standarization of action buttons). What do you think?

javivelasco:
I think they should come from props as you mentioned. In fact, all internationalization in a library such as react-toolbox should be based on that. Anyway there are not a lot of components with text so it should be easy.

sanfilippopablo:
While the props option sounds more technically correct, I think that the
labels being attributes on the locale object is more useful, because the
default locale with the string locale prop is simple option that covers 95%
of the use cases. Besides, it would still allow customization through the
object locale prop.

javivelasco:
Yeah maybe. It also would be a cleaner api. More options on this would be great

PR #681, opened 2016-08-01
Add required polyfills for IE10.

cdroulers:
Two missing polyfills for recent code (I assume) which breaks IE10 compatibility.

In IE10, I still get an error `Unable to get property 'direction' of undefined or null reference` which I'll try to fix in another Pull Request!

### DIFF
--- a/components/utils/polyfills.js
+++ b/components/utils/polyfills.js
@@ -1,5 +1,7 @@
 import 'core-js/fn/array/from';
 import 'core-js/fn/array/iterator';
+import 'core-js/fn/array/find-index';
 import 'core-js/fn/map';
 import 'core-js/fn/string/starts-with';
+import 'core-js/fn/string/includes';
 import 'core-js/fn/symbol';


cdroulers:
I'm not sure if there are tests I can add for IE specifically? Is it possible to run tests with Karma in different browsers? It would be practical to automate regressions or simple usage in different browsers.

cdroulers:
I've found the error in IE10. It seems that using `this.props.whatever` in the default values for the `state` doesn't work!

```javascript
state = {
     direction: this.props.direction,
     focus: false,
     showAllSuggestions: this.props.showSuggestionsWhenValueIsSet,
     query: this.query(this.props.value)
   };
```

I'm not very familiar with babel and its generation yet, but it seems IE10 doesn't scope `_this` properly. Anyone know of a workaround or a way to fix it?

cdroulers:
Any issues? Should I run the build and commit what's in /lib/ as well? Any pointers on the `state` issue?

javivelasco:
Hi @cdroulers! No need to commit the build. In fact that may create more merge issues. I'll do it manually from `dev` branch. Is the block code you've commented still causing issues? If so, we should open an Issue. I'm merging these polyfills. Also, do you think we should better document whose are needed?

cdroulers:
@javivelasco Yes, the code for `state = {` still breaks in IE10 (or 9) and lower. I'll investigate a bit more and open another issue with details. Thank you!

Issue #628, opened 2016-07-11
Ripple effect mostly doesn't work in IE11 (but sometimes does!)


david-cho-lerat-HL2:
Hi!
The ripple effect doesn't work on most components in IE11.
Strangely enough, it works in some places, like in the "hamburger"
IconButton in the "Layout" example on react-toolbox.com.
Why would it work here and not everywhere else ? (buttons, "Ripple"
example, etc). Is it not handled in the same way everywhere ?

Here are three screen grabs from IE11 : 2 not OK and 1 OK

![rippleeffectbuttonie11](https://cloud.githubusercontent.com/assets/19814268/16725913/e69374c4-4759-11e6-8ad8-edae9787b2e4.gif)
![rippleeffectrippleie11](https://cloud.githubusercontent.com/assets/19814268/16725917/ef262988-4759-11e6-9e75-b1b0b126159d.gif)
![rippleeffectlayoutie11](https://cloud.githubusercontent.com/assets/19814268/16725925/f6531b80-4759-11e6-854e-e7da55dc752a.gif)

avdeveloper:
I'm assuming this works on IconButton components as well?

david-cho-lerat-HL2:
Yes, it could be related to mouse coordinates, but it doesn't explain why outiside of `IconButton`s, the ripple effect only shows the last part of the animation : instead of fading in while it's growing, and *then* fading out, it just fades out from full size (see screengrabs from my first post).

david-cho-lerat-HL2:
As stated in my first post, the ripple works correctly with
    IconButtons, as opposed to everywhere else the effect is used,
    as far as I can see.
    
    See the other comments on GitHub.
    
    
    Le 21/07/2016 à 07:03, Albert
      Villaroman a écrit :
    
    
      I'm assuming this works on IconButton components as well? 
      —
        You are receiving this because you authored the thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.

javivelasco:
Surely related to coordinates but it's quite complicated for me to debug it ATM. Help wanted here :)

jcullen:
Per https://developer.mozilla.org/en-US/docs/Web/API/Window/scrollX scrollX/scrollY aren't supported in IE but pageXOffset/pageYOffset work as a cross browser alias. #682

javivelasco:
Thanks for fixing @jcullen

david-cho-lerat-HL2:
[...]

PR #682, opened 2016-08-02
Bugfix for IE scrollX/scrollY

jcullen:
Replaces scrollX and scrollY with cross browser aliases pageXOffset and pageYOffset
Fixes IE issues with ripple and the time picker
#628


javivelasco:
Great! Thanks!

PR #684, opened 2016-08-02
Fixed Autocomplete such that it does not error when provided with a value prop that does not exist in the source prop.

therealparmesh:
Fixed Autocomplete such that it does not error when provided with a value prop that does not exist in the source prop.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -154,7 +154,7 @@ const factory = (Chip, Input) => {
 
    suggestions () {
      let suggest = new Map();
-     const query = this.state.query.toLowerCase().trim() || '';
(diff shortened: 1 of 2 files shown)


therealparmesh:
This problem only occurs for Autocompletes where the provided multiple prop is false.

javivelasco:
I've merged the PR manually to `dev` which is the branch we use for development. `master` is intended to keep the source for the last official release version and PRs should be directed to there.

Thanks for the PR! 💃

zweiler:
I see you've merged already, but is it possible to merge this [PR](https://github.com/react-toolbox/react-toolbox/pull/692) instead? Instead of setting the value to undefined, it preserves the value text.

therealparmesh:
@javivelasco Thank you for merging a fix in for this issue; It really helps in a Redux-driven application. Also, I'll be sure to direct any future contributions to the `dev` branch.
@zweiler Thank you for providing a clean fix!

PR #686, opened 2016-08-04
Fix #685: implement maxLength in javascript instead of using the dom maxlength attribute.

### LINKED ISSUE #685: [Multiline Input Field]: Inconsistency between the character counter and the browser-enforced limit
### The problem

when using the input element with `maxLength` specified, the input field will display the entered amount of characters:

image

The component uses simply the `length` function of the string to determine the amount of entered characters. The limit is enforced using the [`maxlength`](http://www.w3schools.com/tags/att_input_maxlength.asp) attribute of the input. 

However, the maxlength attribute counts newlines (in browsers that implement the specs correct, such as the current version of chrome) as **two characters** (`\r\n`). [Chrome issue](https://bugs.chromium.org/p/chromium/issues/detail?id=252613), [SO discussion #1](https://stackoverflow.com/questions/10030921/chrome-counts-characters-wrong-in-textarea-with-maxlength-attribute), [SO discussion #2](https://stackoverflow.com/questions/22016511/different-maxlength-validation-of-textarea-with-newlines-in-chrome-and-firefox).

To check that out yourself, enter a bunch of newlines [here](https://jsfiddle.net/bxFbL/). For explanations why that's the case, check the [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=252613#c6) in the chrome issue or the [html specs](https://html.spec.whatwg.org/multipage/forms.html#maximum-allowed-value-length).

The `length` function (that is used to display the amount of entered characters, as discussed below) however counts newlines as *one* characters (because in the value string they are just `\n`, the browser just counts them internally as `\r\n` because when the form is submitted that how they are going to be encoded).

That is inconsistent behavior and confuses the users, as the counter displays something like 5/10 but the textarea won't allow you to enter any more characters (this would be the case if you'd enter 5 linebreaks).

### Possible solutions

We should should either adjust the behavior of the textarea limit or the behavior of the counter.

This is a non trival choice - one one hand you want to stick to the specs (and they have a valid point, because if you submit that textarea using a form a newline is submitted as two characters) - but on the other hand that's something the user does not expect - he'd expect that a line break is just one character.

Anyways, to adjust the behavior of the limit, we'd have 

- either to enforce the limit using javascript (which is easy because the input is a *controlled* input, meaning we can just block any input after the `length` function determines that the string reached it maximum length, or better: prevent insertion of characters in the onKeyPress handler)

- or have to let the counter count a newline as two characters, which is trival too. 

I'd prefer the first approach (counting a newline as one character). I could create a PR for this if that'd help.

Cheers,
Simon

SimonSelg:
This follows #685. It fixes the issue by implementing the maxLength restriction in javascript - a newline is now counted as one character.

Also includes multiple small fixes for multiline inputs (see the commit messages).

I needed this for a production app, therefore the rush approach without discussing the solution (as proposed in the issue) first - feel free to reject this PR if you don't like the approach.

Cheers,
Simon

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -111,10 +111,31 @@ const factory = (FontIcon) => {
       this.refs.input.focus();
     }
 
+    onKeyPress = (event) => {
+      // prevent insertion of more characters if we're a multiline input
+      // and maxLength exists
+      const { multiline, maxLength, onKeyPress } = this.props;
+      if (multiline && maxLength) {
+        // check if smt is selected, in which case the newly added charcter would
+        // replace the selected characters, so the length of value doesn't actually
+        // increase.
+        const isReplacing = event.target.selectionEnd - event.target.selectionStart;
+        const value = event.target.value;
+
+        if (!isReplacing && value.length === maxLength) {
+          event.preventDefault();
+          event.stopPropagation();
+          return;
+        }
+      }
+
+      if (onKeyPress) onKeyPress(event);
+    }
+
     render () {
       const { children, disabled, error, floating, hint, icon,
               name, label: labelText, maxLength, multiline, required,
-              theme, type, value, ...others} = this.props;
+              theme, type, value, onKeyPress, ...others} = this.props;
       const length = maxLength && value ? value.length : 0;
       const labelClassName = classnames(theme.label, {[theme.fixed]: !floating});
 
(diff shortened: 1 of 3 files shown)


SimonSelg:
Oh, please don't merge this yet, I thought about this a bit more and noticed some issues with my code. I'll push a fix for them later into this PR.

SimonSelg:
Okay, everything is fixed. Again, the commit messages and comments should explain the changes fairly well.

javivelasco:
This is a great PR! Really good work @SimonSelg , thanks!! 💃

SimonSelg:
Thanks for merging. Any chance you could publish a release soon? We depend on this fix in production, that'd save us having to keep using our fork or the dev branch.

javivelasco:
Sure, I'm releasing today, there are a couple of things more I think I can fix today

PR #687, opened 2016-08-04
Added viewBox attr to svg element of circular ProgressBar to allow custom sizing

### LINKED ISSUE #284: ProgressBar circular is quiet big
The ProgressBar circular version, which is used a lot in Material Design apps, is really big.
As you can see from the Material Design web page (https://www.google.com/design/spec/components/progress-activity.html#progress-activity-types-of-indicators), it should be about 38px.

The one in React Toolbox is about 52px.

The fact is that it's quite hard to change the size of the component (if you use transform: scale() it doesn't look good).

A simple solution could be to have some props like "small" and "big".
So you could have 3 sizes: default (38px), small (24px), big(52px).

And maybe it could also change the stroke-width...

sanfilippopablo:
This should fix #284, I think.

This allows to set a custom size on a circular ProgressBar like so:

```es6
import styles from './styles.scss'
<ProgressBar type='circular' mode='indeterminate' className={styles.progressBar} />
```

```sass
.progressBar {
    width: 32px,
    height: 32px
}
```

### DIFF
--- a/components/progress_bar/ProgressBar.js
+++ b/components/progress_bar/ProgressBar.js
@@ -62,7 +62,7 @@ class ProgressBar extends Component {
 
   renderCircular () {
     return (
-      <svg className={this.props.theme.circle}>
+      <svg className={this.props.theme.circle} viewBox="0 0 60 60">
         <circle className={this.props.theme.path} style={this.circularStyle()} cx='30' cy='30' r='25' />
       </svg>
     );
--- a/lib/progress_bar/ProgressBar.js
+++ b/lib/progress_bar/ProgressBar.js
@@ -73,7 +73,7 @@ var ProgressBar = function (_Component) {
     value: function renderCircular() {
       return _react2.default.createElement(
         'svg',
-        { className: this.props.theme.circle },
+        { className: this.props.theme.circle, viewBox: '0 0 60 60' },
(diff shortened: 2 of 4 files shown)


cdroulers:
Maybe it'd be a good idea to add a test or an example in the `~/spec/components/progress.js` file!

sanfilippopablo:
Totally. I'm on it.

javivelasco:
I'll wait for it!

sanfilippopablo:
OK, guys. I don't know what happened but this time it didn't work as the last time I checked. When applying the styles via `className`, the `width` and `height` properties are overriden by the default ones in `.circular` in the theme. This has two solutions. One is using `!important` and applying it via className. The other one is providing the styles via `theme` and setting width and height on `.circular`.

This is better than as before, when you had to play around with SVG properties to succesfully change size (see #284), while with this one you just change the size of the root element. So this is an improvement anyways.

javivelasco:
The problem was that the custom classes you were giving in the spec had the same priority as the default being the default bundled *after* your custom so it was still taking `60px`. Instead of just passing a `className` I've passed both `className` and `theme` to target the wrapper element and boost priority without using `!important`. That way it works quite well.

Great idea setting `viewBox`, thanks!! 💃

PR #692, opened 2016-08-07
Fixes Autocomplete error when provided value prop that is not in source option while preserving value text

zweiler:
Related to [PR #684](https://github.com/react-toolbox/react-toolbox/pull/684/files) but preserves value text rather than setting empty string.


javivelasco:
Better, thanks!

PR #701, opened 2016-08-10
Fixes for typescript definitions

normano64:
Hi, noticed some problems with the typescript definitions I did. Changed callback functions to Function type as Reacts Event Handlers types may be incompatible as I haven't tested them all. Fixed some typos and wrong type and added every react mouse and touch event.

Also, great work on the project @javivelasco!


javivelasco:
Thanks @normano64 !! Merging!

PR #703, opened 2016-08-10
Added missing modules  'react-addons-update' and 'react-addons-css-transition-group'

ro-savage:
When I tried to build and run the latest react-toolbox docs it failed due to missing `react-addons-update` and `react-addons-css-transition-group`

`lib/tabs/Tabs.js` had been updated at some point but not rebuilt.


mankdev:
Can you release patch with this fix? this breaks projects which uses react-toolbox.

Thanx

javivelasco:
Sure

PR #704, opened 2016-08-10
Added ability to manually input a date into date picker

ro-savage:
This PR adds the ability for the user to manually enter a date if they tab into the Date Picker input field (issue

It accepts the date in the format 'DD/MM/YYYY' or if passed the prop `monthFirst` it will accept 'MM/DD/YYYY. (accepts `/`, `-`, `.` as separators)

If the date isn't in the correct format, the error shows asking for the correct format.
If the date is correctly entered it changes to a proper date object.
If the user clicks onto the input the datepicker dialog opens (a PR is coming for controlling the datepicker by keyboard)

I have only used enzyme for writing tests and wasn't sure how to access a method with `react-addons-test-utils` to write tests for date format checking. However I wrote tests for the regex: http://regexr.com/3e062

Other:
Changed DatePicker and TimePicker to dismiss overlay by default on esc or overlay click.

Happy to hear any feedback.

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -31,6 +31,7 @@ const factory = (Input, DatePickerDialog) => {
       ]),
       maxDate: PropTypes.object,
       minDate: PropTypes.object,
+      monthFirst: PropTypes.bool,
       name: PropTypes.string,
       onChange: PropTypes.func,
       onEscKeyDown: PropTypes.func,
@@ -50,30 +51,28 @@ const factory = (Input, DatePickerDialog) => {
     static defaultProps = {
       active: false,
       locale: 'en',
-      sundayFirstDayOfWeek: false
+      sundayFirstDayOfWeek: false,
+      readonly: false
     };
 
     state = {
-      active: this.props.active
+      active: this.props.active,
+      value: Object.prototype.toString.call(this.props.value) === '[object Date]' ? this.props.value : ''
     };
 
     componentWillReceiveProps (nextProps) {
       if (this.state.active !== nextProps.active) {
         this.setState({ active: nextProps.active });
       }
+
+      if (this.props.value !== nextProps.value) {
+        this.setState({
+          value: Object.prototype.toString.call(nextProps.value) === '[object Date]' ? nextProps.value : ''
+        });
+      }
     }
 
     handleDismiss = () => {
-      this.setState({active: false});
-    };
-
-    handleInputFocus = (event) => {
-      events.pauseEvent(event);
-      this.setState({active: true});
-    };
-
-    handleInputBlur = (event) => {
-      events.pauseEvent(event);
       this.setState({active: false});
     };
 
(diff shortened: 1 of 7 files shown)


ro-savage:
Note #705 should be merged first or else the `onFocus()` & `onBlur()` methods will be added back into `DatePicker` and that will cause the dialog to appear when you tab into the input field.

components/date_picker/DatePicker.js
```diff
@@ -50,33 +51,31 @@ const factory = (Input, DatePickerDialog) => {
...
     };
 
     componentWillReceiveProps (nextProps) {
       if (this.state.active !== nextProps.active) {
         this.setState({ active: nextProps.active });
       }
+
+      if (this.props.value !== nextProps.value) {
+        this.setState({
+          value: Object.prototype.toString.call(nextProps.value) === '[object Date]' ? nextProps.value : ''
```
javivelasco: I think this line can be refactored into a function

components/date_picker/DatePicker.js
```diff
@@ -95,29 +94,79 @@ const factory = (Input, DatePickerDialog) => {
...
+        const seperator = this.getSeparator(_value);
+        const dateSplit = _value.split(seperator);
+
+        return new Date(dateSplit[2], parseInt(dateSplit[1], 10) - 1, dateSplit[0]);
+      }
+
+      return value;
+    }
+
+    isValidDate = (dateString) => {
```
javivelasco: This could be extracted to a `time` method helper along with other methods that are specific to text parsing to a date.

components/date_picker/DatePicker.js
```diff
@@ -95,29 +94,79 @@ const factory = (Input, DatePickerDialog) => {
...
+    }
+
+    getSeparator = (string) => {
+      if (string.indexOf('/') !== -1) return '/';
+      if (string.indexOf('-') !== -1) return '-';
+      if (string.indexOf('.') !== -1) return '.';
+      return '';
+    }
+
+    getErrorMessage = (date, monthFirst) => {
```
javivelasco: To support multilanguage we can move the hardcoded message to a property

javivelasco:
I've left you some comments. BTW where are the tests?? You can test a method retrieving the instance, check the `Slider` tests. Also, you can isolate date parsing methods into a different module and test them from there

javivelasco:
Great work!

javivelasco:
Also, consider it would be nice to toggle the date parsing from a boolean prop so this might be an optional update.

ro-savage:
Will make the changes when I get a chance, sometime next week. Thanks Javi.

components/date_picker/readme.md
```diff
@@ -1,6 +1,6 @@
 # Date Picker
 
-A [dialog](https://www.google.com/design/spec/components/pickers.html#pickers-date-pickers) date  picker is used to select a single date. The selected day is indicated by a filled circle. The current day is indicated by a different color and type weight.
+A [dialog](https://www.google.com/design/spec/components/pickers.html#pickers-date-pickers) date  picker is used to select a single date. The selected day is indicated by a filled circle. The current day is indicated by a different color and type weight. If the input is tabbed into a date can be entered mannual in the format MM/DD/YYYY.
```
leipert: * small typo with manual

leipert:
It's my first time around here, so i hope I do not cause too much trouble ;)
So here is another improvement idea:

Refactor the input parser into an exchangable function, so that a user could provide a function which returns a date. Now a user could leverage for example [momentjs](http://momentjs.com/):

```jsx

// will actually parse '2012 juillet'
const parseDate = (input) => {
	return moment(input, 'YYYY MMM', 'fr').toDate()
}

<DatePicker
        label='Locale (String) - Spanish'
        parseDate={parseDate}
     />

```

ro-savage:
Thanks for the input @leipert 

Seems like a great idea. I probably wont be able to do it as part of this PR but once it's merged would be great if you'd like to contribute it!

javivelasco:
I agree. Great idea @leipert

javivelasco:
Hi @ro-savage !! Are you updating this changes?? It looked very nice :)

ro-savage:
@javivelasco Sorry about that! Between this PR and now, I ended up founding a startup in an accelerator program and work has been non-stop!

I will add this to my calendar and get those changes put through this week.

javivelasco:
No problem! We all have stuff to do outside open source :) Just wanted to know if you have time to keep working on it to either keep or close the PR

ro-savage:
@javivelasco 
I've been out of the loop for a long time now. But I'd like to actually finish this. Should really have done it many many months ago.

Is this still needed?
Should i be merged into dev or different branch now?
Do I need to do something special to make it work for 2.0? (I haven't used React-toolbox since August last year)

javivelasco:
Hi! Well, actually we changes sass to postcss. I don't think it's worthy because I have a new implementation I'm pushing very soon. Maybe we can do it by then

javivelasco:
I'm closing then because there will be too many changes, it's kept to use it as a reference for the future implementation. Also, you'll see you can compose your own! Thanks anyway for the hard work

PR #706, opened 2016-08-12
Support for fixed & inverse color tabs.

Reanmachine:
According to the [Material Design: Tabs Specification](https://material.google.com/components/tabs.html#tabs-types-of-tabs) there is a type of tab called 'fixed' tabs' that evenly distribute their tab headings across the navigation bar:

Fixed Tabs from spec

Additionally there's many examples where the tabs are inverted to match toolbars & whatnot:

Inverse Tabs from spec

This update brings those common spec examples to `react-toolbox` with simple props on the `<Tabs>` element.

React Toolbox Example

### Fixed Tabs Example
```
<Tabs fixed>
   {/* ... */}
</Tabs>
```

### Inverse Tabs Example
```
<Tabs inverse>
   {/* ... *}
</Tabs>
```

They can even be combined.

I think I've gotten all the required pieces around the project, please let me know if I missed anything:
* PropTypes & Theme PropTypes
* Theming Variables in `_config.scss`
* Updated `Readme.md` with example & documentation
* Updated Documentation example
* Updated Spec
* Unit Tests!


javivelasco:
This is great! Thank you for the PR :)

PR #708, opened 2016-08-15
Remove unnecessary spread

cmmartin:
`setState` does a shallow merge, so there is no need to spread the existing state here


cmmartin:
On an unrelated note, we could save these files as `.js` ( to get syntax highlighting, linting, etc) and then just require them using `raw-loader` to make them plain text. We would probably have to tell the `.js` loader to exclude them as well though

```
import checkboxExample from 'raw!./checkbox_example_1.js'
```

javivelasco:
Thanks! This is going to be deprecated very very soon. I'm working on a brand new playground which is going to be *really* awesome. During the next couple of days I'll share a sneak peek :)

cmmartin:
Sweeeeet. Can't wait

PR #710, opened 2016-08-15
Fixes #697 - Autocomplete KeyUp enter fix

### LINKED ISSUE #697: Autocomplete ENTER passing incorrect arguments
Since the version 1.1.0, the Autocomplete.select is updated to use the event but handleQueryKeyUp key 13 is still passing target and event.

jussch:
Fixes issue #697 by giving Autocomplete#select an optional `target` parameter so that the active selection can be passed through from `handleQueryKeyUp()`.

Tested locally, however the build scripts removed all the `.scss` and `.t.ds` files in my environment. Tried to mimic the changes I made with GitHub's editor to avoid that step.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -124,7 +124,7 @@ const factory = (Chip, Input) => {
          }
          this.setState({active: target});
        }
-       this.select(target, event);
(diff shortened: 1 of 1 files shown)


javivelasco:
Thanks! :)

Issue #720, opened 2016-08-22
Tabs component doesn't propagate theme to sub components


ronderksen:
When you create a Tabs component and put a theme property on it, then the theme is not passed on to the Tab and TabContent component. This prevents theme styling of these elements. 
I fixed it by adding a theme property in the renderHeaders and renderContents methods, but I'm not sure that that is the correct way to solve this issue. If you agree that it is, I will create a PR.

javivelasco:
Yeah, we took the approach of giving a single theme for those components that are composed with subcomponents. In this case, since we can inject the theme in the component export, it's possible to forget passing down the theme which definitively should be done.

Your changes sound good, PR it!

ronderksen:
When do you expect to push a new release to npm?

javivelasco:
Mmm maybe a change affected in the good way 😎 
A new release should come during the week, I just wanted to review some pendings PRs to release it all together :)

PR #726, opened 2016-08-23
Theme tab sub components

ronderksen:
As discussed in issue #720, this is the PR to solve that issue.


javivelasco:
Great! Thanks 💃

PR #735, opened 2016-08-24
Passes IconMenu's theme to the inner Menu.

jussch:
Small change.

Tried to theme the IconMenu and found that theme wasn't being passed through. I assume that handing IconMenu's theme to the Menu inside won't cause errors.

### DIFF
--- a/components/menu/IconMenu.js
+++ b/components/menu/IconMenu.js
@@ -71,6 +71,7 @@ const factory = (IconButton, Menu) => {
             ripple={this.props.menuRipple}
             selectable={this.props.selectable}
             selected={this.props.selected}
+            theme={this.props.theme}
           >
             {this.props.children}
           </Menu>


javivelasco:
Thanks! It won't generate problems. This is not going to be an issue anymore, themr has now namespacing in case you want to use a module to target a different component and there are class collisions 😁

jussch:
Great! 

To clarify, are you saying that I can namespace the theme object I pass to IconMenu to target the Menu underneath? If so, that sounds mighty useful! Is it documented anywhere yet? I'd love to take a look, thanks.

javivelasco:
Yeah, exactly. We published a release yesterday but the documentation still needs to be updated. You can find an example of usage in the test folder and [here](https://github.com/javivelasco/react-css-themr/pull/14) meanwhile. Still needs to be integrated in react-toolbox but it looks like that's covering the only issue I have until now with themr!! 💃

PR #736, opened 2016-08-25
add fullscreen dialog.

acornejo:
A first version of fullscreen dialogs as defined in the material spec
https://material.google.com/components/dialogs.html#dialogs-full-screen-dialogs


javivelasco:
Thanks! 👍

PR #737, opened 2016-08-25
navbar hide on scroll behavior.

acornejo:
Add basic hide on scroll behavior for navbar. As described in

https://material.google.com/patterns/scrolling-techniques.html#scrolling-techniques-app-bar-scrollable-regions

The diff looks larger because I had to transform the component into a class to keep some state. However the actual changes are quite small. Namely we add a scrollHide class when the user scrolls down "enough" (where enough is the height of the navbar), and we remove the scrollHide class as soon as the user starts scrolling up. The actual animation is done through css transitions.

### DIFF
--- a/components/app_bar/AppBar.js
+++ b/components/app_bar/AppBar.js
@@ -1,67 +1,95 @@
 import React, { PropTypes } from 'react';
+import ReactDOM from 'react-dom';
 import classnames from 'classnames';
 import { themr } from 'react-css-themr';
 import { APP_BAR } from '../identifiers.js';
 import InjectIconButton from '../button/IconButton.js';
 
 const factory = (IconButton) => {
-  const AppBar = ({ children, leftIcon, onLeftIconClick, onRightIconClick, rightIcon, theme, title, ...props }) => {
-    const className = classnames(theme.appBar, {
-      [theme.fixed]: props.fixed,
-      [theme.flat]: props.flat
-    }, props.className);
-
-    return (
-      <header className={className} data-react-toolbox='app-bar'>
-        {leftIcon && <IconButton
-          inverse
-          className={classnames(theme.leftIcon)}
-          onClick={onLeftIconClick}
-          icon={leftIcon} />
-        }
-        {title && <h1 className={classnames(theme.title)}>{title}</h1>}
-        {children}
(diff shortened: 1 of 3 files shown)


javivelasco:
Nice! Let me check it tomorrow. Thanks for all PRs you're doing lately

acornejo:
Great, thanks for this toolbox!.

Btw, I've also thought about this change a bit more, and now I am leaning towards an alternative implementation which keeps track of the scroll state in a separate component and simply add a prop to the navbar to do the hiding animation.

That way we could reuse the scroll part if we wanted to add hide-on-scroll behavior to the bottom navigation component or the tabs. Keep that in mind when you look at this, I look forward to your comments.

javivelasco:
Let me know about the progress. There are tons of changes I want to make, just lacking time to do so lately. One of those is to move a lot of stuff to HOC. The scroll control feels like it can be one of those :D

javivelasco:
Sorry @acornejo I've been very busy and totally forget about this great PR. Can you rebase so we can merge just today?

acornejo:
rebased @javivelasco, more work than i expected.

how is the postcss branch going? I've postponed contributing more components to avoid churn.

acornejo:
(the travis-ci checks are failing due to lint errors in files I didn't touch in this commit.. which is odd to say the least)

javivelasco:
@acornejo I plan to merge all pending before the next weekend and to release a `2.0.0-beta.0` with the PostCSS changes. The branch is fully usable except from some steps that can be taken during the build process (just to avoid unnecessary postcss plugins) and the documentation.

javivelasco:
I think it's failing because of linting errors. If you rebase again it should pass

acornejo:
Thanks Javi, rebasing again did the trick. I'll be looking forward to the beta release, I plan to start contributing back as soon as it is out. Thanks for your hard work!

javivelasco:
Merged! Great work @acornejo 

Just changed a couple of things:
- Only attach the event listener if AppBar is hideable. Had to check the nextProps on component will receive props lifecycle.
- Remove ReactDOM dependency by using `ref` callback property which I think is more performant.

Thanks for your work!

PR #738, opened 2016-08-25
Introducing BottomNavigator

Kikobeats:
Hello,

I'm trying to write this component, that's my current status.
### Markup

```
<BottomNavigation active={1}>
  <BottomNavigationItem icon='home' label='Home' />
  <BottomNavigationItem icon='announcement' label='Messages' />
  <BottomNavigationItem icon='account_box' label='Me' />
</BottomNavigation >
```
### TODO
- [ ] Implement change between views.
- [ ] Implement optional  _badge_ prop.
- [ ] Extract style in variables.
- [ ] Add docs.
### Discuss: Integrate it in Tabs?

Personally I detect that this component is very similar to tabs except in:
- It uses `label + icon` instead of `label` for the navigation item.
- In tabs, the active view it places under the navigation and bottomNavigator place the active view above the navigation list. It the same just change the order!
- By definition, this component have location limitations. For example, could be possible put bottomNavigator on top? like [Twitter mobile webapp](https://mobile.twitter.com/home) follow a similar approach but placing in in the top.

screen shot 2016-08-25 at 09 20 14

So could be perfectly reasonable extend Tabs for support this approach.

I need feedback and help to finish that, go ahead! 😄

### DIFF
--- a/components/bottom_navigator/BottomNatigator.js
+++ b/components/bottom_navigator/BottomNatigator.js
@@ -1,0 +1,32 @@
+import React from 'react';
+import style from './style';
+import FontIcon from '../font_icon';
+import classnames from 'classnames';
+
+class BottomNavigationItem extends React.Component {
+  handleClick = (event) => {
+    this.props.onClick(event, this.props.index);
+  };
+
+  render () {
+    const { active, label, index, icon, theme, ...other } = this.props;
+    const classes = classnames(style.li, {
+      [style.liActive]: active
+    });
+
+    return (
+    <li
+      {...other}
+      data-react-toolbox='bottom-navigation-item'
+      className={classes}
+      onClick={this.handleClick}
+    >
+      {this.props.children}
+      <FontIcon className={style.icon} value={icon} />
+      <span className={style.span}>{label}</span>
+    </li>
+    );
+  }
+}
+
+export default BottomNavigationItem;
--- a/components/bottom_navigator/BottomNavigatorItem.js
+++ b/components/bottom_navigator/BottomNavigatorItem.js
@@ -1,0 +1,46 @@
+import React, { PropTypes, Component } from 'react';
+import style from './style';
+
+class BottomNavigation extends Component {
+  static propTypes = {
+    active: PropTypes.number,
+    children: PropTypes.node
+  };
+
+  static defaultProps = {
+    active: 0
+  };
+
+  state = {
+    active: this.props.active
+  };
+
+  componentWillReceiveProps (nextProps) {
+    if (this.props.active !== nextProps.active && this.state.active !== nextProps.active) {
+      this.setState({ active: nextProps.active });
+    }
+  }
+
+  render () {
+    const { active: defaultActive, children, theme, ...other } = this.props;
+    const { active } = this.state;
+    const filteredChildren = React.Children.map(children, (child, idx) => (
+      React.cloneElement(child, {
+        active: active === idx,
+        index: idx,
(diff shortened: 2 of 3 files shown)


acornejo:
Hi Kikobeats,

This is great! I was about to start working on this when I noticed your PR!. Can't want to see it merged. 

I would not make the views part of the bottom navigation component. In fact, I would like to refactor the tab component to stop doing that. I just opened a discussion about it in issue #739, I would love to hear your thoughts (and Javi's) on the matter. 

Instead I would suggest we implement the "change between views" functionality on a separate component; we could then reuse this for tabs, bottom navigation, and even when switching views when clicking on a sidepanel, or a navigation bar, etc.

Second, while your current screenshot looks great, the bottom navigation bar is out of proportion (too large). The spec says it should have 56dp height, same as the appbar. See here https://material.google.com/components/bottom-navigation.html#bottom-navigation-specs

Finally, while I agree there are a lot of similarities between the tabs and bottom navigation, there are also a few important differences. One difference is the styling of the labels all caps for tabs, and Capital case for bottom navigation. Another difference is the way in which the active item is indicated, where tabs use a styled underline, and bottom navigation uses either icon color (the one you implemented), or icon color combined with an animated hiding of the labels in the inactive components. A video of this is available in the link I provided.

So while sharing code is great, I fear that by merging these two components we will end up with something which is harder to read and maintain. (Part of the beauty of react-toolbox is that the code is very well structured, whereas other toolkits *cough caugh material-ui* provide 101 props per component and have super complex logic.

Of course, in the end its all up to you and Javi ;)

Kikobeats:
Thanks for the feedback @acornejo!

The intention of this PR is not be merged yet, just is a inflexion of point for determinate how code this component and do it from the scratch (instead of open an issue and suggest code one 😄 ).

I think that Tabs logic component is very complicated right now, extract change between views in something reusable definitely sounds a correct step to do!

Reanmachine:
@Kikobeats - Would we not be served better by expanding the `Tabs` control to support icons, anchor position. I recently had a PR published that allowed the tabs to have fixed-width regions.

This is where the the material design docs leave me a bit conflicted about code re-use. Ultimately the docs specify a tabs variant that has the icons.

https://material.google.com/components/tabs.html#tabs-usage

So what we have are Fixed-Width Tabs, with Icons, anchored to the bottom of the tab content instead of the top. However, just to make things difficult they have an acually specified component:

https://material.google.com/components/bottom-navigation.html

The only thing that's different is the behavior similar to apple's app bar with the "zoomed" active button and the smaller icons.

I'm wondering if improving the Tab Control to have most of the behaviors we want here is ideal then make the BottomNavigation configure the tab in a predictable way and `flex: 1` the Tab Content so it fills it's parent.

@javivelasco May have some thoughts on component direction & re-use. Also keep in mind he's working on the `css-modules` transition for v2 and you may want to take that into account when making this component.

javivelasco:
@Reanmachine sorry I'm not paying attention to this PR ATM. I'm trying to keep up to date with issues and stuff and it's quite time consuming. I'm not considering the addition of more components until the full migration to PostCSS. To make the process less painful I'm releasing a beta version, coming during next days

owais:
It makes sense for the community to release new components as independent packages. This one could be `react-toolbox-bottom-navigation` or `rt-components-bottom-navigation`. react-toolbox core team can then select the best 3rd party components to include in core as they mature.

javivelasco:
This is really outdated. I'm closing, maybe in the next stable release it can be included

PR #740, opened 2016-08-25
add tab icons.

acornejo:
This adds the possibility of specifying an icon to use in the tab
headers. You can specify either a label only, icon only or both.

https://material.google.com/components/tabs.html#tabs-usage

### DIFF
--- a/components/tabs/Tab.js
+++ b/components/tabs/Tab.js
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
+import { FontIcon } from '../font_icon';
 import { themr } from 'react-css-themr';
 import { TABS } from '../identifiers.js';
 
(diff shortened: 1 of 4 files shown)


components/tabs/Tab.js
```diff
@@ -40,6 +42,18 @@ class Tab extends Component {
     }
   };
 
+  renderIcon = () => {
+    if (typeof this.props.icon === 'undefined') {
+      return null;
+    } else if (typeof this.props.icon === 'string') {
+      return (
```
javivelasco: Is this div really needed? Also I think handling the string is done inside the FontIcon component so maybe it's not needed here. Did you tried??

acornejo: The div is needed to ensure the tab icon and the tab label appear on separate lines as required by the spec (without it they appear side-by-side).

You are right about the string handling in FontIcon, so I fixed that.

Issue #745, opened 2016-08-27
Replace onMouseDown with onClick in Dropdown


laucheukhim:
Is there a reason `onMouseDown` is used instead of `onClick` in dropdown menus?

MouseDown event is not fired reliably in Cordova and thus the Dropdown component does not work there.

Can we use `onClick` instead (PR: #746)?

PR #746, opened 2016-08-27
Replace onMouseDown wth onClick in Dropdown

laucheukhim:
See #745.


javivelasco:
It looks ok but if we were using it there should be a reason 😕 Since it looks like it works the same I'm merging!

PR #749, opened 2016-08-29
Add app_bar title, leftIcon and rightIcon props.

acornejo:
Here we add some basic default functionality for the most common use
cases of the AppBar component. It is still possible to provide and style
your own elements through children props, but for most use cases the
props will help reduce the amount of boilerplate.

### DIFF
--- a/components/app_bar/index.d.ts
+++ b/components/app_bar/index.d.ts
@@ -13,6 +13,18 @@ export interface AppBarTheme {
    * Added to the root element when the app bar is flat.
    */
   flat?: string;
+  /**
+   * Used as the app bar title.
+   */
+  title ?: string;
+  /**
+   * Added to the left icon app bar element.
+   */
+  leftIcon?: string;
+  /**
+   * Added to the right icon app bar element.
+   */
+  rightIcon?: string;
 }
 
 interface AppBarProps extends __ReactToolbox.Props {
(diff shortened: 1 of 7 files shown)


components/app_bar/index.d.ts
```diff
@@ -31,6 +43,26 @@ interface AppBarProps extends __ReactToolbox.Props {
    */
   flat?: boolean;
   /**
+   * If it exists it is used as the AppBar title
+   */
+  title? string;
```
normano64: Hi, you probably just missed it but there should be a `:` between the property and type ( `title?: string`), like you did above.

acornejo: thanks! fixed.

javivelasco:
Noticed I've made a change in this PR before merging. The `AppBar` no has a dependency with `FontIcon`. Since `FontIcon` has no style dependencies it shouldn't be an issue but I've followed the same practice we are applying in the whole library about dependency injection.

In react-toolbox you can import a component without styles so they can get styles from the context of from a theme property. This is ok for most component but if a particular component has dependencies with other, the dependency should come in the same fancy. This means that importing a dependency component directly from the same definition would resolve it always the same way: with of without styles. What we do to avoid this is you create a `factory` for the component receiving the dependencies. Then, at the bottom of each file we expose the component with the themr applied but raw and the factory. Then, in the `index.js` we are importing and injecting styles.

Take this into account in case you need to add dependencies to any component!
Thanks for the PR! :D

mmazzarolo:
Nice feature :)
Any reason behind the use of `FontIcon` instead of `IconButton `?

acornejo:
Good question. The new appbar icon props accept both strings or elements, so you could use an iconbutton if you wanted.

As far as i can tell the only difference is that iconbutton has a ripple effect by default (icon button provides a couple of other props, but none would be used). Ripple seems like a good idea for the appbar, so it may be a better choice than fonticon anyway.

@javivelasco do you want a pr to switch the appbar from fonticon to iconbutton?

javivelasco:
Yeah, it fits better!

PR #751, opened 2016-08-30
Adding hideMode to Tabs

### LINKED ISSUE #709: Need option to mount tabs that are not active
I have a use case where it would be more convenient if hidden tabs were mounted (but still hidden). Some of the tabs we're showing make AJAX calls to fetch data and it's expected that users will be switching between the tabs. So having the tabs mounted and hidden makes the UX better as the user can immediately see what they are expecting when they click on a tab. 

I feel the default should be the current behavior. If there is agreement on this feature and it's config I can put together a pull request.

I think it should be configured on `Tabs` and not an individual `Tab`. I can think of two ways to configure this:

1. `hideMode: PropTypes.oneOf(["display", "unmounted"])`. default: "unmounted"
1. `mountHidden: PropTypes.bool`. default: false

What do you think?

abhirao:
- Defaulting to `unmounted` - not mounting inactive tabs. `display` mode will mount inactive tabs but keep them hidden.
- Closes #709

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -12,6 +12,7 @@ const factory = (Tab, TabContent) => {
       className: PropTypes.string,
       disableAnimatedBottomBorder: PropTypes.bool,
       fixed: PropTypes.bool,
+      hideMode: PropTypes.oneOf(['display', 'unmounted']),
       index: PropTypes.number,
       inverse: PropTypes.bool,
       onChange: PropTypes.func,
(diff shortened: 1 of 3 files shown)


abhirao:
This is my first PR to this project. This will address #709. Please let me know if you want me to adjust anything. 

I wasn't sure about how to update the docs for the Tabs so I manually added the new prop to the Tabs' readme. If you think it's worth an example I can include that.

Thanks!

javivelasco:
Great!! I love when you guys contribute with tests, it's something we need to do with the lib. Great work, thank you!

PR #752, opened 2016-08-31
Fixed errors for Typescript 2.0.2

normano64:
The latest RC of Typescript gave errors for having file extensions in the imports.
Should still work with older versions of Typescript but haven't tried anything older than the 2.0.0 beta.


javivelasco:
Thanks!

Issue #476, opened 2016-04-22
Table renders incorrect with normalize.css@4.1.1


farwayer:
![2016-04-22-134400_1312x241_scrot](https://cloud.githubusercontent.com/assets/2343101/14739468/5bece0b0-0890-11e6-8699-e0ea4c230433.png)

With normalize.css@3.0.3 all ok:
![2016-04-22-134334_1113x222_scrot](https://cloud.githubusercontent.com/assets/2343101/14739471/5ed6ce26-0890-11e6-9dd0-ce5702647f67.png)

langri-sha:
Maybe this #446 is related to your issue?

langri-sha:
What version did you upgrade from?

farwayer:
`0.16` (and upper) version of `react-toolbox` peer depends on `normalize.css@4` and with it table renders incorrect (no row delimiters and vertical white gaps on selected row).

farwayer:
And something wrong with date input also.
![2016-04-24-060034_726x252_scrot](https://cloud.githubusercontent.com/assets/2343101/14765241/06fc6898-09e2-11e6-9bd1-2dd95f752d82.png)

Tested on firefox and chrome.

javivelasco:
Gonna check it during the weekend!

javivelasco:
Updated dependency!

farwayer:
@javivelasco issue is still exists with `normalize.css@4`. You can see it in project table component example.

javivelasco:
Thanks! Gonna check on it

farwayer:
Fixed by #758. Thanks @rubenmoya!

javivelasco:
Thanks for closing!

PR #758, opened 2016-09-01
Fix table borders (Issue #476)

rubenmoya:
Table borders were shown incorrectly because of the `border-collapse: separate` property (Check https://github.com/react-toolbox/react-toolbox/issues/476) .

I changed the `border-spacing` to 0 and moved the border to the cells instead of the row.

Feedback is welcome :D

Before:
screen shot 2016-09-01 at 12 23 07

After:

screen shot 2016-09-01 at 12 22 28

### DIFF
--- a/components/table/theme.scss
+++ b/components/table/theme.scss
@@ -8,10 +8,10 @@
   font-size: $font-size-tiny;
   color: $table-text-color;
   text-align: left;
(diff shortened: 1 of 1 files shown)


javivelasco:
Thanks @rubenmoya !

PR #763, opened 2016-09-02
Update components/list/index.d.ts

deniss-muhla:
Wrong type for ListSubHeaderProps.caption. Should be string instead of boolean.


javivelasco:
Thanks!

PR #764, opened 2016-09-02
Make Table multi-selectable again

tuckerconnelly:
Currently the table isn't multi-selectable (check out http://react-toolbox.com/#/components/table and try to select multiple rows).  This patch fixes that.


tuckerconnelly:
Hmmm something happened with the build--the relevant file is `components/Table/Table.js` :)

PR #771, opened 2016-09-07
fix active year style in date picker.

acornejo:
The active year font-size didn't have the units set (which means it
didn't work). Also to better match the material design spec the
font-weight for active years is set to semi-bold.


javivelasco:
Thanks!

PR #775, opened 2016-09-07
Pass down theme from Card title to nested Avatar

aleksanb:
The current approach to theme the nested avatar was by adding a `div:first-child` selector in the .cardTitle class.
A bit less than ergonomic.


javivelasco:
Thanks!

PR #778, opened 2016-09-07
Replace onMouseDown wth onClick in DatePicker and TimePicker

laucheukhim:
Continuing on #745, I am replacing `onMouseDown` with `onClick` on DatePicker and TimePicker.

These two components do not work properly when running in Cordova. The DatePickerDialog and TimePickerDialog do not show up when you tap on the Input component of DatePicker and TimePicker because the mousedown event is not fired reliably.

There are other components that use `onMouseDown` and they probably don't work properly in Cordova. I will create future PRs as I come across them.


therealparmesh:
@laucheukhim Awesome fix, thank you!

laucheukhim:
@javivelasco Any updates on this PR? Is there something I need to fix before it can be merged?

javivelasco:
Sorry, I couldn't check it out. I'm not sure if there is something that might not work or if it was done like that due to something that is already changed.

I'm merging, this should come with the next patch (weekend!)

PR #779, opened 2016-09-07
Fixing issue where closing an overlay on top of another overlay re-enables document body scrolling unnecessarily.

therealparmesh:
I'm having an issue regarding scrolling on the document `body` when an overlay is active.

**Steps to reproduce:**

1. Have a document `body` that will scroll.

2. Open a `<Dialog />` that contains a `<DatePicker />`.

3. Open the `<DatePicker />`.

4. Close the `<DatePicker />`.

**Observed behavior:**

Despite the `<Dialog />` still being open and the document `body` still being grayed out behind it, the document `body` regains its scrolling ability.

**Expected behavior:**

The document `body` should not be scrollable until all overlays are inactive.

**Proposed solution:**

Check for other overlays before blanking out the `overflow` style attribute on the document `body`. Please let me know if this is a good approach.

### DIFF
--- a/components/overlay/Overlay.js
+++ b/components/overlay/Overlay.js
@@ -33,7 +33,7 @@ class Overlay extends Component {
 
   componentWillUpdate (nextProps) {
     if (nextProps.active && !this.props.active) document.body.style.overflow = 'hidden';
-    if (!nextProps.active && this.props.active) document.body.style.overflow = '';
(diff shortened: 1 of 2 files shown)


javivelasco:
Thanks for the PR! I think the solution is appropriated and right now I'm not sure how can it be improved. It does the trick so may you solve conflicts? Then we can merge :)

therealparmesh:
Understood! Will do.

therealparmesh:
Should be good to go.

javivelasco:
Thanks!

PR #784, opened 2016-09-09
appbar: fonticon -> iconbutton

acornejo:
now we get a nice ripple effect on the appbar icons.

this factory thing to get ripples going is a bit annoying. What would folks say about a slightly different effect (sacrificing the ripple origin) achieved with css only?

here is an initial thought:

http://codepen.io/hastyrobot/pen/XjmAQL

### DIFF
--- a/components/app_bar/AppBar.js
+++ b/components/app_bar/AppBar.js
@@ -2,9 +2,9 @@ import React, { PropTypes } from 'react';
 import classnames from 'classnames';
 import { themr } from 'react-css-themr';
 import { APP_BAR } from '../identifiers.js';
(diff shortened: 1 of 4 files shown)


javivelasco:
Thanks @acornejo Actually the ripple thing is a bit tricky. It should be kept until mouse up, track the origin, expand multiple times with multiple clicks... The factory is not used just for the ripple but to allow to get components with dependencies with/without styles.

For example, if you want to import a raw `Button` component, it should not require styles for the Ripple either and viceversa. The factory allows to inject the dependency component forgetting about styles. I know it may be annoying at first but we export injected components too so requiring the factory is optional from outside. More difficult but way more powerful too :)

PR #785, opened 2016-09-09
Use contentEditable for editable table cells

tuckerconnelly:
This makes it so editable text cells in Table wrap, instead of getting cut off.

screen shot 2016-09-09 at 2 24 37 pm

### DIFF
--- a/components/table/TableRow.js
+++ b/components/table/TableRow.js
@@ -75,26 +75,14 @@ const factory = (Checkbox) => {
       const inputType = utils.inputTypeForPrototype(this.props.model[key].type);
       const inputValue = utils.prepareValueForInput(value, inputType);
       const checked = inputType === 'checkbox' && value ? true : null;
-
-      if (inputType === 'text') {
-        return (
-          <div
-            contentEditable
-            suppressContentEditableWarning
-            onInput={this.handleInputChange.bind(null, index, key, inputType)}>
-            {inputValue}
-          </div>
-        );
-      } else {
-        return (
-          <input
-            checked={checked}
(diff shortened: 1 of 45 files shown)


acornejo:
I was interested in this, it was hard to read because your commit touches 45 files, I don't think you need to include the lib/* files as part of your PR.

coljung:
I think you editing all the lib/* files make this merge not good. If you are updating table cell in the Table component, those should be the only files in your PR.

javivelasco:
I agree. Please update the changes removing the build so it can be more clear to review. In any case a new build should be created after the release / merge with master!

tuckerconnelly:
Okay updated the PR. Some revert conflict weirdness happened so one lib file is still in there, but should be easier to read now.

Why is the lib even committed? Isn't .npmignore enough?

components/table/TableRow.js
```diff
@@ -62,14 +75,26 @@ const factory = (Checkbox) => {
...
+      if (inputType === 'text') {
+        return (
+          <div
+            contentEditable
+            suppressContentEditableWarning
+            onInput={this.handleInputChange.bind(null, index, key, inputType)}>
+            {inputValue}
+          </div>
+        );
+      } else {
```
coljung: I dont think the else > return is necessary in this case. See: http://eslint.org/docs/rules/no-else-return

javivelasco: +1 Apart from that great!

tuckerconnelly: Aight it's in

PR #793, opened 2016-09-14
Add onBlur and onFocus props in Autocomplete component

ducreyna:
Add missing props functions onBlur and onFocus for Autocomplete component.
That's really useful because Autocomplete is based on Input component.


javivelasco:
Thanks!

PR #795, opened 2016-09-14
Correct docs for CardTitle component

stianjensen:
`large` is applied when the CardTitle has no avatar, not the other way around.


javivelasco:
Thanks!

PR #798, opened 2016-09-15
Remove overlay: hidden from the body tag when an overlay is closed in IE11

unculture:
In IE11 setting

```document.body.style.overflow = null;```

does not work. But

``` document.body.style.overflow = "";```

does.

Using the empty string works for me today in Chrome Canary, Firefox Dev, iOS Safari, OSX Safari 9.1.1.


unculture:
Yeah. I'll close and reopen with single quotes.

PR #799, opened 2016-09-16
Fixes scroll after overlay closing in IE11

unculture:
In IE11 setting

```document.body.style.overflow = null;```

does not work. But

``` document.body.style.overflow = '';```

does.

Using the empty string works for me today in Chrome Canary, Firefox Dev, iOS Safari, OSX Safari 9.1.1.


unculture:
I've changed to single quotes to fit the style guide.

javivelasco:
Great, thank you!

PR #812, opened 2016-09-22
Add table model header details to docs

chiester:


javivelasco:
Great! Thank you!

I'm thinking from a while ago that this implementation should be changed. It is indeed handly for most cases but if you want to do a very specific customization you just can't. Probably we can modify the table to support nesting as usual with other componentes and keep this kind of "ModelTable" to generate tables quickly. The other option should be supported thought. What do you think?

PR #818, opened 2016-09-25
Removed semvar lock from travis ci node versions

amilajack:


javivelasco:
Retest this please

javivelasco:
It looks like tests are not passing with this build!

amilajack:
Can you try rerunning them?

javivelasco:
It was a problem due to a duplicated default export I think. This shouldn't make any trouble so merging! Thanks!

PR #821, opened 2016-09-26
Use material icons for radio state

### LINKED ISSUE #797: Decentered radio button
Depending on the zoom level of the browser, I can see the center of the checked radio button misaligned:
radio-buttons

I tried on different browsers (Chrome, Safari, Firefox) all with the same problem.

Is there any reason not to use the Material icons for this purpose?
  * https://design.google.com/icons/#ic_radio_button_checked
  * https://design.google.com/icons/#ic_radio_button_unchecked

yhbrandon:
Resolves #797.

This uses material icons instead of generating radios with borders. The animation ending up being a bit different. Also adjusted the size to match material spec

### DIFF
--- a/components/date_picker/index.js
+++ b/components/date_picker/index.js
@@ -1,5 +1,3 @@
-export default from './DatePicker';
-
 import { themr } from 'react-css-themr';
 import { DATE_PICKER, DIALOG } from '../identifiers.js';
 import { datePickerFactory } from './DatePicker.js';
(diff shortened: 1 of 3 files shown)


javivelasco:
I'm afraid this will not be accepted, sorry @yhbrandon and thanks for the PR. The reason is that the radio should support an animation (that was broken BTW but I already fixed). With icons the animation is just not possible and there is an small delay when switching because the content has to be changed and that is more costly that just animating an internal pseudoelement.

PR #823, opened 2016-09-27
Apply Checkbox style property to internal .check

### LINKED ISSUE #791: Override checkbox color
Theming components in react-toolbox is easy using css and css modules, but sometimes a one-off styling is required. Say, you have a list of checkboxes and want to style each of them in separate colors.

Currently, you would need to create a class for each color you want to set and then apply that as a className for the root Checkbox component. But since the checkbox itself is nested within and can't be controlled directly, you need to specify it with nested selectors:

```
.colorRed .check {
    background-color: red;
    border-color: red;
}

<Checkbox className={theme.colorRed} theme={theme} />
```

Even though inline styling is bad for general situations, in these cases inline colors are useful to override specific elements and avoid creating a bunch of specific classes (especially when colors are generated in js based on data).
It is possible to set the style attribute of the Checkbox element, but all those styles are currently applied to the internal input element, which is not what is seen as the colored checkbox.

Would it be possible to for instance add a color attribute to the Checkbox component to allow for just this situation? The value of the color attribute could then be applied as inline styling on the internal Check component as background and border color.

I fully agree with this project's decision to not use inline styling, and that is the exact reason why I moved from material-ui to react-toolbox, but I still believe that inline styles can be useful in cases like what I've described. If you agree that it could be useful or have suggestions for alternatives, I'd be happy to help out with a pull request.

stianjensen:
This will allow you to override the styles of the visible check element
instead of the hidden input element.

This fixes #791.

### DIFF
--- a/components/checkbox/Check.js
+++ b/components/checkbox/Check.js
@@ -2,11 +2,12 @@ import React, { PropTypes } from 'react';
 import classnames from 'classnames';
 
 const factory = (ripple) => {
-  const Check = ({checked, children, onMouseDown, theme}) => (
+  const Check = ({checked, children, onMouseDown, theme, style}) => (
     <div
       data-react-toolbox='check'
       className={classnames(theme.check, { [theme.checked]: checked })}
       onMouseDown={onMouseDown}
+      style={style}
     >
       {children}
     </div>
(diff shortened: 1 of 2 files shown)


javivelasco:
Thank you!

PR #827, opened 2016-09-28
Updated Sass/ Typescript build commands to add Windows support and added support for clicking on row in table

ryanmcafee:
- Added support for sass and typescript copying of files for Windows users, cpx requires double quotes and not single quotes

- Added the ability to click on a row via passing down a onClick handler

- Removed a duplicate default export from the datepicker file caught in the linting process


javivelasco:
Thanks

PR #832, opened 2016-10-01
add table header width and table cell render function

q3452:
This incorporates the [enhancements](https://github.com/joehua87/react-toolbox/commit/fbc7d2b6237782eaea3899ea549aa47b7c217e5b) introduced by [joehua87](https://github.com/joehua87) that never seemed to become a PR.  I've slightly updated them to fit the current react-toolbox code which seems to have slightly progressed from the version Joe forked, e.g. his title is no longer needed.

### DIFF
--- a/components/date_picker/index.js
+++ b/components/date_picker/index.js
@@ -1,5 +1,3 @@
-export default from './DatePicker';
-
 import { themr } from 'react-css-themr';
 import { DATE_PICKER, DIALOG } from '../identifiers.js';
 import { datePickerFactory } from './DatePicker.js';
--- a/components/table/TableHead.js
+++ b/components/table/TableHead.js
@@ -5,8 +5,8 @@ const factory = (Checkbox) => {
     let selectCell;
     const contentCells = Object.keys(model).map((key) => {
       const name = model[key].title || key;
-      const width = model[key].width;
-      return <th key={key} style={{width}}>{name}</th>;
+      const headStyle = model[key].headStyle;
+      return <th key={key} style={headStyle}>{name}</th>;
     });
 
     if (selectable && multiSelectable) {
--- a/components/table/TableHead.js
+++ b/components/table/TableHead.js
@@ -5,7 +5,8 @@ const factory = (Checkbox) => {
     let selectCell;
     const contentCells = Object.keys(model).map((key) => {
       const name = model[key].title || key;
-      return <th key={key}>{name}</th>;
+      const width = model[key].width;
+      return <th key={key} style={{width}}>{name}</th>;
     });
 
     if (selectable && multiSelectable) {
--- a/components/table/TableRow.js
+++ b/components/table/TableRow.js
@@ -55,8 +55,13 @@ const factory = (Checkbox) => {
         return this.renderInput(key, value);
       } else if (value) {
         if (this.props.model[key].render) {
-          const disabled = true;
-          return this.props.model[key].render(this.props.index, key, value, disabled);
+          return this.props.model[key].render(
+            {
+              index: this.props.index, 
+              key: key, 
+              value: value, 
+              disabled: true,
+            });
         }
         return value.toString();
       }
@@ -66,8 +71,13 @@ const factory = (Checkbox) => {
       const index = this.props.index;
       const inputType = utils.inputTypeForPrototype(this.props.model[key].type);
       if (this.props.model[key].render) {
-          const disabled = false;
-          return this.props.model[key].render(index, key, value, disabled);
+          return this.props.model[key].render(
+            {
+              index: index, 
+              key: key, 
+              value: value, 
+              disabled: false,
+            });
       }
       const inputValue = utils.prepareValueForInput(value, inputType);
       const checked = inputType === 'checkbox' && value ? true : null;
--- a/components/table/TableRow.js
+++ b/components/table/TableRow.js
@@ -38,7 +38,8 @@ const factory = (Checkbox) => {
 
     renderCells () {
       return Object.keys(this.props.model).map((key) => {
-        return <td key={key}>{this.renderCell(key)}</td>;
+        const style = this.props.model[key].style;
+        return <td key={key} style={{style}}>{this.renderCell(key)}</td>;
       });
     }
 
--- a/components/table/TableRow.js
+++ b/components/table/TableRow.js
@@ -54,7 +54,7 @@ const factory = (Checkbox) => {
         return this.renderInput(key, value);
       } else if (value) {
         if (this.props.model[key].render) {
-          var disabled = true;
+          const disabled = true;
           return this.props.model[key].render(this.props.index, key, value, disabled);
         }
         return value.toString();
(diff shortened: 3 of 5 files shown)


q3452:
It broke on ThemedDatePicker, I don't think I caused this.

```
components/date_picker/DatePicker.js -> lib/date_picker/DatePicker.js
components/date_picker/DatePickerDialog.js -> lib/date_picker/DatePickerDialog.js
components/date_picker/__test__/index.spec.js -> lib/date_picker/__test__/index.spec.js
SyntaxError: components/date_picker/index.js: Only one default export allowed per module. (19:0)
  17 | 
  18 | const ThemedDatePicker = themr(DATE_PICKER, theme)(DatePicker);
> 19 | export default ThemedDatePicker;
     | ^
```

components/table/TableHead.js
```diff
@@ -5,7 +5,8 @@ const factory = (Checkbox) => {
     let selectCell;
     const contentCells = Object.keys(model).map((key) => {
       const name = model[key].title || key;
-      return <th key={key}>{name}</th>;
+      const width = model[key].width;
+      return <th key={key} style={{width}}>{name}</th>;
```
javivelasco: I think is way more handy if instead of using a `width` you use a `style` tag. Anyway, you are applying a style from the model to the Head. Not sure if it should be applied in the body too. Or, in case it doesn't, it should be `headStyle` and `style`.

In any case, this implementation is probably going to change quite soon. It is not really flexible if you want to do very specific customizations.

components/table/TableRow.js
```diff
@@ -60,6 +60,9 @@ const factory = (Checkbox) => {
     renderInput (key, value) {
       const index = this.props.index;
       const inputType = utils.inputTypeForPrototype(this.props.model[key].type);
+      if (this.props.model[key].render) {
+        return this.props.model[key].render(this.props.data);
```
javivelasco: I don't get this. Is it like a method to notify that rendered had happened? If so, naming it `onRender` would me more appropriated IMO

components/date_picker/index.js
```diff
@@ -1,5 +1,3 @@
-export default from './DatePicker';
```
javivelasco: If you rebase with `dev` this should not be a problem. In any case we have to pull it out of the PR

components/table/TableRow.js
q3452: No, it calls a render function supplied in the data model, the test for props.model[key].render is checking for the existence of a render function in the model, if it finds one then it's used to render the table cell and further rending is prevented by early return.

I'm implementing code from @joehua87 but I think his intent was reasonably clear and it seems to work pretty well.

This is in answer to #268 which describes the problem and this solution more detail but essentially, the model becomes:

`const model={ someColumn: {type: String, render: function(data){ return (MyDropDown(data)) } } };`
Where MyDropDown is defined elsewhere in scope:
```
const dropDownData = {...};
const MyDropDown = function(data) { return (<Dropdown
        auto
        label={'Choose data item'}
        onChange={this.handleMDDChange}
        source={dropDownData}
        value={this.state.source.dropDownValue} />
```)};

This lets the dev embed arbitrary controls into the table cell and use the control's onChange to keep them synchronised with the table state.

File: components/table/TableHead.js (hunk 1 above)
Comment (q3452) [inline components/table/TableHead.js]: It **is** using a style tag, the width is just the name of the variable in the data model.  Maybe widthStyle would be better.  It should definitely be applied to the body too, I can change that.

File: components/date_picker/index.js (hunk 3 above)
Comment (q3452) [inline components/date_picker/index.js]: Will do.

File: components/table/TableHead.js (hunk 1 above)
Comment (javivelasco) [inline components/table/TableHead.js]: I know it's coming from the model, but I mean that it could be more general if you just use `style` instead of with so the model defines a full style object. This way you could customize the whole header cell not just width.

File: components/table/TableRow.js (hunk 2 above)
Comment (javivelasco) [inline components/table/TableRow.js]: Alright. Anyway I want to let you know I'd rather change this implementation to another one closer to material-ui. At the first implementation I wasn't sure about the best way to compose this kind of components but after a year it's pretty clear that this approach is not flexible enough in some cases.

**Comment by javivelasco** [comment]
Fix #268

**Comment by q3452** [comment]
I found a bug that complicates the implementation a bit.
The custom control also needs to handle disabled state (when table has no onChange), which means its render function needs a {disabled} parameter, otherwise a locked table has an editable dropdown.  Also, in order to update the table's datasource, the control needs to pass its row/column to the parent's onChange callback.

I'll rebase, change 'width' to style, apply that style to relevant table body columns, add the extra parameters...then see what you think of the PR.

**Comment by q3452** [comment]
Table body columns are now styled using a `style` attribute on the model entry.  Table Headers are styled with `headStyle` attribute on the model entry.  A render function can be placed on the model entry to render the table cell.  This render function should return an appropriate control for the data type.  It must look like `render(row, col, value, disabled)`, the row&col are used to update table data on the parent element via callback if required `source[row][col] = value` and the `disabled` parameter should be obvious.

**Comment by q3452** [comment]
How does it look now?  If you approve of those changes, I'll see about providing some documentation for the render function.

File: components/table/TableHead.js [hunk 4]
```diff
@@ -5,8 +5,8 @@ const factory = (Checkbox) => {
     let selectCell;
     const contentCells = Object.keys(model).map((key) => {
       const name = model[key].title || key;
-      const width = model[key].width;
-      return <th key={key} style={{width}}>{name}</th>;
+      const headStyle = model[key].headStyle;
+      return <th key={key} style={headStyle}>{name}</th>;
```
javivelasco: That's what I mean 👌

components/table/TableRow.js
```diff
@@ -53,6 +54,10 @@ const factory = (Checkbox) => {
       if (onChange) {
         return this.renderInput(key, value);
       } else if (value) {
+        if (this.props.model[key].render) {
+          const disabled = true;
```
javivelasco: I'm ok with this but I just have a small code suggestion. I never like to pass boolean arguments because of the semantics and I see that to make it clear you assigned here a variable meaning what that third parameter.

Since now we have kind of named parameters with objects and destructuring, what do you think about passing an object instead?

```js
this.props.model[key].render({ 
  index: this.props.index,
  disabled: true
  key
});
```

q3452: It needs value as well but sure, then the parameter order is irrelevant.  You're right that the variable assignment was purely to make the code self-documenting.
```
this.props.model[key].render({ 
  index: this.props.index,
  disabled: true,
  key,
  value,
});
```
To be honest, I want the change in first so that I could use your playground to write an example that I know works and then I could push that as the new playground example code.

javivelasco: Great

q3452:
Are we able to pull this now?  Once it's in and you update your docs site, I can use that to write a working example and do another PR for that.

javivelasco:
Please check the new implementation. I think it's no longer needed but I really appreciate you worked on this! Thank you! 💯

PR #843, opened 2016-10-05
Fixed typo in list/config.css

marcuscarrigan:


lib/list/config.css
```diff
@@ -8,7 +8,7 @@
   --list-divider-height: calc(0.1 * var(--unit));
   --list-item-min-height: calc(4.8 * var(--unit));
   --list-item-min-height-legend: calc(7.2 * var(--unit));
-  --list-item-hover-color: $palette-grey-200;
+  --list-item-hover-color: palette-grey-200;
```
rubenmoya: I think this should be `var(--palette-grey-200)` to work.

Issue #626, opened 2016-07-10
Future components / wishlist


javivelasco:
I'd like to list in this issue components that are wanted but not implemented yet. I'd distinguish two kind of components: defined by the spec but not included/not defined by the spec, but useful and implemented by other frameworks.

### From the spec
- [ ] `FAB` as discussed [here](https://github.com/react-toolbox/react-toolbox/issues/517)

### Out of the spec
- [ ] `BottomSheet` Implemented by Angular Material [here](https://material.angularjs.org/latest/demo/bottomSheet)
- [ ] `Timeline`defined by @dlebedynskyi [here](https://gist.github.com/dlebedynskyi/64a28f366cdd41d6ab01e957d4a0d755)

I want to use this thread to discuss if components out of the spec should be created in a separated repository (react to this post with 👍) or inside the same repository under a different folder (react with 👎 ). I'm more with the first approach. Also the list can be used to track the progress and to hold references to other issues and PRs where the components are developed.

Consider this kind of a *roadmap* of new components. Make your requests replying to this instead of opening new issues.

vmsoftware:
@javivelasco 

Really a great idea. I think RT is the best MD library that I have seen never and congratulations for it. great work!. 

However I think the actual implementation of MD need some changes that would be excellent before think about new components or external components, I mean **responsive components**, a **12-columns based grid system** to organize the content and a **theme switcher** with 'light theme' and 'dark theme', proposed by MD specification.

I think those changes are more important at this moment and I'm going to suggest my personal opinion about how it should be:

For responsive components we only need to set the width of each component external wrapper to 100% and organize the internal content relative to this wrapper. You can say it's not usefull, but if you think about 12-columns based grid system really is the best option. Why?

let me try to explain it,

The 12-columns grid system, work as follow: 

1. Make a `.row` class with 100% of width.

2. Take 100% and divide between 12 then each columns have 8,3333333333% of width.

3. Make CSS classes for 1,2,3,4,5,6,7,8,9,10,11 and 12 columns like col1, col2, col3.... col12.  (8,3333333333% * column number). for example 

4. Make mediaQuerys for `.row` and `.colxx` classes if necessary.

```
.col3 {
 width: 25%;
}
```

Then the components with the 100%, fill `.colxx` wrapper and organize the components easy.

It looks like a hard work but if we take advantage of existing libraries it's really easy. 

[...]

PR #844, opened 2016-10-05
+ Pager control, actually it is not defined at matereal design but de…

MaximKomlev:
Pager control, actually it is not defined at matereal design but definitely it is helpful to have.

the control is based on react-toolbox controls and corresponds to matereal design. It is highly customisable and it follows default style patterns of react-toolbox components. actually what differentiate the control from others similar that it has always const set of buttons and the set depends on two properties: number of total pages (latsPage) and number of visible pages at block. e.g.
[<][1][...][4][5][6][...][8][>] or [<][1][2][3][4][5][...][8][>] or [<][1][...][4][5][6][7][8][>] where block size is 3. it gives smooth user experience. A layouting is based on flex but easily can be rewritten using inline-block at custom theme.

example is added to specs.


javivelasco:
Great work! The code looks very good and respects React Toolbox patterns, lovely ❤️ 

There is an open discussion #626 about components that are out of the spec though. We agreed that maybe the best option is to create a separate repository for this complementary components and I agree that would be probably the best to separate official vs opinionated components.

Also, a PostCSS release is coming so maybe you'd like to update. I think it could be cool to add a section in the docs mentioning complementary components. This would keep this component in a separated repository with a dependency to react-toolbox. When we put together some components with can open a "react toolbox addons package".

Not sure yet anyway but in any case I'd prefer to merge to PostCSS directly.
What do you think??

javivelasco:
Also check out the linter errors!

MaximKomlev:
Hi Javi,
What you are saying it makes sense, i think so. I didn't find any related repository or project therefore I've tried to publish my experiments such way. Definitely, it would be great to have repository where people could share his experience around react-toolbox to make it more powerful. I found react-toolbox very useful and highly customizable, I like it.

neekey:
@MaximKomlev good work, need this component too. Any plan about how to publish your component?

javivelasco:
@MaximKomlev thank you so much. If you want, you can publish your component and open a PR with a modification in the Readme or Docs with a new section detailing the link to your component. Something like `Complementary components` or `Components out of the spec`... something like that.

If the component is not included in this same repository cause it's more opinionated, we still can include a link to your repo until we set up a complements repository. What do you think??

MaximKomlev:
@javivelasco I will do this way thank you.

javivelasco:
Thank you for an awesome work @MaximKomlev ! Please don't forget that PR to the react toolbox docs/readme

MaximKomlev:
@javivelasco @neekey  Hey, guys I started project for additions which I guess could be helpful to have but which are not included at react-toolbox components library.
https://github.com/MaximKomlev/react-toolbox-additions
I would so appreciate if you could give me your review before publishing.

neekey:
@MaximKomlev Nice work, I think I will take a look tonight : )

neekey:
@MaximKomlev I haven't gone through all your implementation details only read about your API, and I think it's not very clear:

- `prevButtonContent` and `nextButtonContent`: Since their prop types are `string` I think renaming them like `nextButtonLabel` will be easier to understand
- `lastPage`: I was quite confused by this one at first, after going through your code I then realized it is actually `totalPage`, which is a more conventional name.
- `rangeLeftButtonContent`: First of all, it's quite a new feature to me that you make the "range" clickable, but I'm not sure if this is necessary. Secondly, it's very hard to understand what is "rangeLeftButton" at the first glimpse, and still, may consider replacing `content` to `label`.
- `***ButtonTypes`: I noticed the prop type is `object` but the name is `*Types`, quite confused me. I suggest instead of giving the control of all separated buttons, you can just provide a whole component level theme prop to set predefined button style set like:

```js
<Pager raised  ... />
<Pager flat  ... />
``` 
Since I don't think anyone will set left button to raised and right button to flat.

Good work 👍 by the way.

MaximKomlev:
Thank you, for your review, I updated interface according to your comments. actually left and right buttons for ranges and arrow have one style, just can be populated by different content. I attached screen of how it can be styled.

I did clickable the ranges button such I think it more useful.
pager_screenshot

neekey:
@MaximKomlev the screenshot looks good 👍

MaximKomlev:
Hey, I published first version of rt addons, for now, I added two components Pager and FilePicker, if it will be interesting to somebody I have a couple more ideas about components.
https://www.npmjs.com/package/react-toolbox-additions

javivelasco:
@MaximKomlev please PR a README update detailing the project link. Other people can find it useful

MaximKomlev:
I will... on this weekend, thank you.

MaximKomlev:
@javivelasco , Hi, I would like to confirm..., so it should be something like:
##Extensions
....

Before
## License
is it ok?
Thank you.

javivelasco:
Sure, yeah!

Issue #780, opened 2016-09-07
Tooltip TypeError when quickly hovering between multiple tooltips


gregegan:
When you move the cursor really quick between the tooltip elements an error is thrown in console.
`Uncaught TypeError: Cannot read property 'style' of undefined`

![untitled](https://cloud.githubusercontent.com/assets/6665100/18331737/133aa83c-7516-11e6-9d95-1779dd25c224.gif)

https://github.com/react-toolbox/react-toolbox/blob/dev/components/tooltip/Tooltip.js#L84 

![image](https://cloud.githubusercontent.com/assets/6665100/18331799/7e112e38-7516-11e6-80a8-6465b7d0c6b6.png)

gregegan:
Can anyone else validate this bug?

vitalis:
Same here!!

juneidysoo:
same thing happened, if the Tooltip-ed button throws an error, the button's onClick wouldn't work

javivelasco:
Fixed!

PR #846, opened 2016-10-07
Fixed Tooltip TypeError when quickly hovering between multiple tooltips

rrott:
This PR is for issue https://github.com/react-toolbox/react-toolbox/issues/780 "Tooltip TypeError when quickly hovering between multiple tooltips"

When you move cursor between elements too fast, the `transitionEventNamesFor()` function could be called with an element that does not exist anymore so  `element.style[transition]` throw the error.


javivelasco:
Great! Thanks! :)

PR #847, opened 2016-10-07
File browse button

MaximKomlev:
the component just slightly extends current button and provide file picker functionality, actually the button can be merged to original button with a specifier like type|mode.


javivelasco:
Great work, this should be definitively done but I have one suggestion. Don't you think it would be better to include this  along with the `Input` component? For example, right now you can use the `Input` component passing a `file` type property. In that case, the component is not looking good because we don't take into account that case.

I think we can be abstracted from the component used to represent a file browse and instead just create an input. Internally, the `Input` can just check the passed type and when it's an type file, you just render this component. This way `FileBrowse` would be a subcomponent of `input` that it's not directly accessed by the user but by react-toolbox.

MaximKomlev:
Actually yes, you are right, I thought so and mentioned about that at the comment (but I told about Button component not about Input) that it is better to add type where e.g. file could be specified (why button, because from my perspective the component looks like button and it is expectable to have it there). But at the same time consumer of lib (developer) by using input component with type of file will expect to have filepicker component. So at my perspective it is good to have filebrowse button component and extend Input component and returns filepicker component when at Input component type is file, it gives more flexible way to use. What do you think?

javivelasco:
I get your point but still think we can use this implementation directly inside the `Input`. If a developer sets a Input with type set to file she will get an undesired output anyway because right now it's not considered. Using this instead of nothing is definitively a  step forward. Anyway we can keep it as a Button and use it from the Input so ok. Let's merge and we can set it up afterwards! Thanks!

Issue #766, opened 2016-09-05
Input hint text should appear by default not just focused


neekey:
Is this the default behavior that hint only appears when focused? 

It is actually quite weird since hint should be used as a "hint" if it does not appear by default, how can a user get the hint at the first glance of the form?

Material-UI's TextFileld will show hint by default: http://www.material-ui.com/#/components/text-field

Not focused:

![image](https://cloud.githubusercontent.com/assets/499870/18237843/a41460b2-7379-11e6-94df-20242f918272.png)

Focused:

![image](https://cloud.githubusercontent.com/assets/499870/18237845/ac8ae3ce-7379-11e6-905f-5a69ae87f5e5.png)

My own user case (which was made with Material-UI):

![image](https://cloud.githubusercontent.com/assets/499870/18237854/d1c24f88-7379-11e6-9dac-e156d9bb5fab.png)

![image](https://cloud.githubusercontent.com/assets/499870/18237862/db050be4-7379-11e6-9867-8cf348ac61b8.png)

javivelasco:
This is happening because usually you have a label and in those cases the hit appear with focus. Actually there is a bug because if there is no label the hint should appear directly. Fixing asap

rowanG077:
Maybe it would be better to autofloat the label if a hint is available. Or add a prop to always float the label.

javivelasco:
This is fixed!

Amir-Mousavi:
Please fix this issue for password filed too.

PR #848, opened 2016-10-09
Updated to show hint on input if label is empty

rrott:
Updated to show hint if label is missed (issue: https://github.com/react-toolbox/react-toolbox/issues/766)
20u8e1nfmb


javivelasco:
Thanks!

Issue #253, opened 2016-01-08
CardMedia is not getting shown in Firefox


YoYuUm:
Sometimes a CardMedia component with an image in a Card component doesn't show the image at all in Firefox, some other times it shows only the image partially. 

The problem is related with the "padding-top" rule set using percentages. But I didn't figure out what's exactly the problem. You can use as example the component in react-toolbox documentation: http://react-toolbox.com/#/components/card.

I'm using Firefox 43.0.4 under os X.

javivelasco:
Thanks, I'll investigate it

nathanmarks:
@YoYuUm @javivelasco 

I think I know the fix for this -- we may need an additional container. The `padding-top` media area property is a css-only "hack" for forcing the aspect ratio.

Check out the answer on this SO question: http://stackoverflow.com/questions/23717953/padding-bottom-top-in-flexbox-layout

derekstavis:
I was about to report the same issue here, which I noticed when playing with playground. In my code, adding an additional container worked for me too. Thanks @nathanmarks :)

javivelasco:
Alright!

nathanmarks:
@javivelasco the container should be added to the lib IMO

javivelasco:
mmmm ok

nathanmarks:
Yeah, FF has a huge % of browser share still, can't ignore it :)

derekstavis:
At office most people have recently migrated from Chrome to Firefox because of resource usage. Some of my friends and even costumers also reported the same trend. Although most browser historical data shows firefox usage declining, the [press reports different behviours](http://gizmodo.com/fuck-it-im-going-back-to-firefox-1685425815).

javivelasco:
Of course, we have to provide fully support

PR #849, opened 2016-10-09
Added ::after pseudo to CardMedia to fix an issue in FireFox

rrott:
Added ::after pseudo to CardMedia to fix the issue with not showing images in FF
https://github.com/react-toolbox/react-toolbox/issues/253

FireFox:
screen shot 2016-10-09 at 21 17 12

Chrome:
screen shot 2016-10-09 at 21 28 59


javivelasco:
Thanks!! 💃

PR #852, opened 2016-10-11
fixes old sass var in css config

vectorsize:
There was an old sass variable in the config.css file that breaks for css-only setups.


javivelasco:
Thanks!

Issue #584, opened 2016-06-22
Add required to Dropdown as Input fields


suancarloj:
Hello,

Currently there is no way to indicate that a dropdown is required like the input fields per example.

Thanks!

javivelasco:
As easy as adding a as spread a prop to the inner `Input` component. PR welcome for this!

PR #857, opened 2016-10-11
Added required to Dropdown

rrott:
Added required property to indicate that a dropdown is required
https://github.com/react-toolbox/react-toolbox/issues/584

screen shot 2016-10-11 at 23 06 29

screen shot 2016-10-11 at 23 06 50

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -20,6 +20,7 @@ const factory = (Input) => {
       onChange: PropTypes.func,
       onClick: PropTypes.func,
       onFocus: PropTypes.func,
+      required: PropTypes.bool,
       source: PropTypes.array.isRequired,
       template: PropTypes.func,
       theme: PropTypes.shape({
@@ -30,6 +31,7 @@ const factory = (Input) => {
         errored: PropTypes.string,
         field: PropTypes.string,
         label: PropTypes.string,
+        required: PropTypes.bool,
         selected: PropTypes.string,
         templateValue: PropTypes.string,
         up: PropTypes.string,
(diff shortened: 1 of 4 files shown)


javivelasco:
Good one!

components/dropdown/Dropdown.js
```diff
@@ -157,6 +167,7 @@ const factory = (Input) => {
             {...others}
             className={theme.value}
             onClick={this.handleClick}
+            required={this.props.required}
             readOnly
```
jimmywarting: I wanted to validate a form with html5 constraint validation while setting the required property. 

Applying the `readonly` treats the input as valid, making the required attribute pointless.

I  want to get rid of the readonly attribute

context: https://stackoverflow.com/questions/12777751/html-required-readonly-input-in-form

dsullivan7:
I'm using `2.0.0-beta.12` and I've noticed that the `required` attribute does not function as expected:

```
class DropdownTest extends React.Component {
  state = { value: null }

  handleChange = (value) => {
    this.setState({value: value})
  }

  render() {
    return (
      <Dropdown
        required
        onChange={this.handleChange}
        source={[{value: 1, label: 'option 1' }, {value: 2, label: 'option 2' }]}
        value={this.state.value}
      />
    )
  }
}

const Form = () => (
  <form>
    <DropdownTest />
    <Button type="submit" raised primary>Submit</Button>
  </form>
)
```

In this example, the submit button still executes the default submit function. I was able to use the `Input` component with the `required` attribute and get the desired functionality.

jimmywarting:
Should create an issue about it

dsullivan7:
@jimmywarting thanks, will do. For reference the opened issue is: #1789

PR #860, opened 2016-10-12
Fix Overlay Memory Leak

HaNdTriX:
The [Overlay Component](https://github.com/react-toolbox/react-toolbox/blob/dev/components/overlay/Overlay.js) doesn't remove the `keydown` listener when it is unmounted.
This is because `this.handleEscKey.bind(this) !== this.handleEscKey` [(code)](https://github.com/react-toolbox/react-toolbox/blob/dev/components/overlay/Overlay.js#L29)

This prevents garbage collection of the component as well fires the `handleEscKey` listener even when the component has been unmounted.

@javivelasco Since this is a critical bug I would love to see this fix in Version 1 as well.

### DIFF
--- a/components/overlay/Overlay.js
+++ b/components/overlay/Overlay.js
@@ -26,7 +26,7 @@ class Overlay extends Component {
 
   componentDidMount () {
     if (this.props.active) {
-      this.escKeyListener = document.body.addEventListener('keydown', this.handleEscKey.bind(this));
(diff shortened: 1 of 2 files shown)


javivelasco:
Of course, I'm releasing a patch now. Thanks for the fix!

HaNdTriX:
wow that was fast. Thanks a lot!

javivelasco:
No problem! :)

PR #863, opened 2016-10-13
Remove normalize.css from commons.css

razor-x:
This splits out global styles defined in commons.css into global.css.
This gives users the option to include either commons.css
(just as before) or global.css (which is everything but normalize.css).

Users should have the option to handle importing external dependencies
like normalize.css separately from the rest of the toolbox's common styles.

This also includes a commit to fixup da0c470.

### DIFF
--- a/components/commons.css
+++ b/components/commons.css
@@ -1,4 +1,3 @@
-@import 'normalize.css';
 @import './colors.css';
 @import './variables.css';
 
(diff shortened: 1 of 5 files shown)


javivelasco:
I wonder if we should just remove normalize encourage people to include their own.
Wouldn't be better?

razor-x:
Yes, that's my preference as well. Most projects will use a css reset anyway. When third party libs bundle them it creates version conflicts and duplication. 

I can update this PR to remove it instead. The docs should probably mention that normalize is recommended.

razor-x:
@javivelasco Updated.

PR #864, opened 2016-10-13
add support for FontAwesome

neerfri:
react-toolbox rocks!

I was missing font-awesome, which providers a lot of icons missing from the material icon pack.

This is an attempt at making react-toolbox support font-awesome using a simple icon name prefix.
It's early work, so there's probably a lot of things that can be improved

WDYT?

### DIFF
--- a/components/font_icon/FontAwesomeIcon.js
+++ b/components/font_icon/FontAwesomeIcon.js
@@ -1,0 +1,21 @@
+import React, { PropTypes } from 'react';
+import classnames from 'classnames';
+
+const FontAwesomeIcon = ({name, modifiers, className, ...rest}) => {
+  const modifierClasses = modifiers ? modifiers.split(',').map((m) => (`fa-${m}`)) : null;
+  return (
+    <span
+      className={classnames(className, 'fa', `fa-${name}`, modifierClasses)}
+      {...rest}
+    />
+  );
+};
+
+
+FontAwesomeIcon.propTypes = {
+  className: PropTypes.className,
+  modifiers: PropTypes.string,
+  name: PropTypes.string
+};
+
+export default FontAwesomeIcon;
--- a/components/font_icon/FontIcon.js
+++ b/components/font_icon/FontIcon.js
@@ -1,16 +1,27 @@
 import React, { PropTypes } from 'react';
-import classnames from 'classnames';
+import MaterialIcon from './MaterialIcon';
+import FontAwesomeIcon from './FontAwesomeIcon';
 
-const FontIcon = ({ children, className, value, ...other}) => (
-  <span
-    data-react-toolbox='font-icon'
-    className={classnames({'material-icons': typeof value === 'string' || typeof children === 'string'}, className)}
-    {...other}
-  >
-    {value}
-    {children}
-  </span>
-);
+const IconSpecRegExp = /(?:([^:]+):|)([^:]+)(?::(.*)|)/;
+
+const FontIcon = ({ value, ...rest}) => {
+  if (typeof value === 'string') {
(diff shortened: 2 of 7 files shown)


javivelasco:
Thanks! Great work, not sure if font-awesome makes sense here having MDI, mostly because you can use a custom Icon component that uses FontAwesome right now. The dependencies are injected if you check the source so every component has a factory function where you can inject its dependencies. Those components using an Icon that needs a FontAwesome icon can be created from the factory function injecting your custom.

I would love to get more feedback before merging!

neerfri:
@javivelasco Thanks for your response!

Using the factories was my first thought, but it had 2 problems. 1) it would require to re-define a lot of components and 2) it would not allow to use both MDI and FA (without using a component that can do both like I do here).

Another option that exists is to always use the font component itself, for example I can do `<Button icon={ <FontAwesome name="some-icon" /> } />`
It's not so bad, just requires more code and more imports in a lot of files.

In general I think the right approach is to use the `icon` prop as a component because every application will probably need it's own icon set(s) eventually. regarding FA I thought it might be useful to have it in the core since it is very popular and MDI is currently lacking (IMO).
If we decide to go forward it would make it much better if I add some documentation before (or after) we merge.

The bad thing about this PR is that it might communicate that this is "the way" to integrate more font packs, and it probably shouldn't. Maybe adding a separate section in the documentation about using custom font packs (showing the `icon` prop usage) will solve that.

So there are pros and cons to merging this. As you said, it's probably best to get some more input from other maintainers as well.

Best,

Zequez:
It doesn't need to FontAwesome specific. You could call it `ClassFontIcon` and it could turn anything like `icon:cog:spin` into `icon icon-cog icon-spin`, and it would support any font library, like Fontello.

And since we're at it we could make a generic `LigatureFontIcon`, although I think the Material font is the only popular ligature-based font available right now.

Zequez:
I'm thinking maybe something like this:

`home` -> `<span class="material-icons">home</span>`
`other:home` -> `<span class="other">home</span>`
`other.spin:home` -> `<span class="other other-spin">home</span>`
`fa.home` -> `<span class="fa fa-home"></span>`
`icon.home` -> `<span class="icon icon-home"></span>`
`fa.cog.spin:text` -> `<span class="fa fa-cog fa-spin">text</span>`

javivelasco:
Agree, I would make a very generic approach so you can easily integrate other fonts.

advance512:
Any update?

neerfri:
As I wrote in: https://github.com/react-toolbox/react-toolbox/pull/864#issuecomment-255658069
I think this is not the right way to solve the problem.
I think using `<Button icon={ <FontAwesome name="some-icon" /> } />` is.

** Why? **
1. there a re a few ways to implement fonts (including awesome) in react (global css, css modules, ...). using the global option by default is not very wise nor future-proof.
2. react-toolbox's  codebase should not be concerned with the fonts you use. it should remain composable.
3. with good JS practices the code overhead can be very small: `<Button icon={ i("heart") } />` is not much longer than `<Button icon="heart" />`.
4. this method will make the codebase of the app using this library much more flexible. you can implement a function to provide your icon system-wide and use your own semantic names for them in the rest of your code.

As such I'm closing this PR.
Best,

javivelasco:
@neerfri Sorry for leaving this that long. For the future version we are probably removing the `icon` property from the button so you should nest the Icon using children. I think it's way more flexible like that :) I appreciate so much your PR anyway, thank you :)

neerfri:
Software gets better with every PR merged, and even more so with every PR not merged 😃 
Thanks for you're awesome work on the toolbox!

Issue #870, opened 2016-10-14
May 1st missing


k0r8i:
Hi there!
Did anyone encounter May 1st not showing up in languages different to English?
I can reach the date by using cursors though it's not visible.
(Screenshot taken from react-toolbox.com attached)

Cheers
Korbi
![maymissing](https://cloud.githubusercontent.com/assets/17675159/19391659/e1b0a872-922d-11e6-85f1-ecde40b39c41.png)

rrott:
The problem is in calculating correct position for the first day of the first month's week.
In this particular May the very first day starts on Sunday and for countries where weeks start on Sunday, it's the first day of the first May's week while for others it's the last day of the first May's week. 

I am adding PR that fixes this issue.

PR #871, opened 2016-10-15
Fixed issue with rendering calendar incorrectly when first day of a m…

rrott:
This PR is a fix for https://github.com/react-toolbox/react-toolbox/issues/870 - May 1st missing

The problem is in calculating correct position for the first day of the first month's week.
If it's on Sunday and Sunday is the first day of the week, current calculation is correct, but in case Sunday is the last day of the week, it sets -1 as the position and we don't see this day.  

screen shot 2016-10-15 at 14 00 41

screen shot 2016-10-15 at 14 00 52


javivelasco:
Great, thanks!

PR #873, opened 2016-10-16
Add optional cancel handler in dialog dismiss

pepesenaris:
Allow optional handler for onDismiss event of Dialog for DatePicker components. If the active state is managed outside the DatePicker you have no way setting active false when closing the dialog


components/date_picker/DatePicker.js
```diff
@@ -134,8 +135,8 @@ const factory = (Input, DatePickerDialog) => {
             maxDate={maxDate}
             minDate={minDate}
             name={this.props.name}
+            onDismiss={onDismiss || this.handleDismiss}
```
javivelasco: I would rather use the `handleDismiss` always and then call `onDismiss` from the same handler in case it exists. Looks more clear if the handler is always called.

PR #878, opened 2016-10-18
ts def typo fix and addtions

panjiesw:
This enables us typescript users to use such component properties.
Also, I use react-toolbox heavily with typescript at work, so I'll likely create this kind of pull request again. :smiley:


javivelasco:
Thanks!

PR #881, opened 2016-10-19
New Table implementation

### LINKED ISSUE #237: Table component should support onFocus and onBlur prop
I would find it very useful to pass an onFocus and onBlur prop to the ```<Table />``` component.  Ideally, the callback fires whenever a cell of the table is focused or blurred.

### LINKED ISSUE #268: Ability to add buttons to a Table

### LINKED ISSUE #531: Add some features to Table
Can you add 3 things to table?

1. Ability to select a row when you click on whole row.

2. Ability to hide checkboxes but can still select rows.

3. Ability to add onClick handler to each rows.

### LINKED ISSUE #543: Table allows react elements as cell values but that is not clear from the documentation
Allowing table cell values to be react elements is great flexibility. However this is not clear in documentation. 

I suggest updating the documentation to include this important information.

### LINKED ISSUE #796: How to add custom field to Table component
It seems you can add custom ( React components ) to the Table one.

For example, I have my source which is 5 fields, and i would like to add a custom one at the end with a button/link to edit such row.

So far i havent been able to achieve this, but i think its possible to have this minor customization.

thanks!

### LINKED ISSUE #825: Add Button to Table
Hi, btw react-toolbox is awesome.
But,  I wanna ask something. Can I add button in react-toolbox Table?, like add table in HTML table like this way:
`<table>
<tbody>
<tr>
<td><button>Click this!</button>
</tr>
</tbody>
</table>`

And one more question, how to merge two or more cell like `<td colspan="3">Column</td>` in HTML table?
Thank you.

### LINKED ISSUE #830: Table column widths
It'd be nice to be able to specify the widths of table columns. Maybe just allowing a `children` prop to Table that's just added before the table head and body, where you could have e.g. `<colgroup> ... </colgroup>`.

### LINKED ISSUE #880: A new approach for Table
The `Table` component is not flexible enough. Some people might find useful to set a model to inline edit fields but the reality is that's really limited. It should be reimplemented following a different approach.

EDIT: Please check the PR or leave your feedback here!

javivelasco:
The `Table` component is quite a weirdo. I'm sure a lot of people might find useful having a model and fields that are directly editable but the truth is that it's quite limited. I've just wrote a new implementation which is closer to Material Design spec. Here are some benefits:

- Instead of one `Table` accepting a model and rendering subcomponents, now you will compose your table using `Table`, `TableHead`, `TableRow` and `TableCell`.
- You can use a property to set a cell column as sortable so it will be shown with an arrow.
- The state is extracted but the API is kept. You will still receive an array with indexes of the selected elements but this state is not externally stored and it's your responsibility to render selected rows appropriately. 

Now you can do easily stuff that was pretty difficult like having a sorted table, pagination, etc.
Please I'd ❤️ to get some feedback before merging, leave yours! Also I'd need somebody to write typings for this. Maybe @normano64 or @panjiesw ?

Fixes #880 
Fixes #830 
Fixes #825
Fixes #796 
Fixes #543 
Fixes #531 
Fixes #268 
Fixes #237 

Here is how it looks:

screenshot 2016-10-19 02 12 21


normano64:
I think this will be quite useful 👍 
I can write typings for it this weekend and make a PR with some other fixes.

panjiesw:
This will indeed be tremendously useful :+1: 
It is more like the spec described about Table component.

I ended up creating my own table component with extra sort and filter, but with these new implementation, I can just extend/use the head, row, and cell component inside some custom container. It's more flexible.

And yes I can help with the typings. But, regarding ts definitions, I think there are also other areas I'd like to improve, if you're okay with that. I can see from the code that you pass along other props to the native component, for example the react-toolbox `Input` component. Ideally, since it passes other props to the native `input` element, the ts definition of `InputProps` should extends the one provided by react's, `React.HTMLProps<HTMLInputElement>`

coljung:
Looks great, cant wait to give it a try.
I will take a look functionality wise once i can play with it.

javivelasco:
Ok I think it definitively worthy to merge, thank you all! @panjiesw any work on TS is welcome, I'm not a user and can't keep everything fine and tested on my own even though I know it's not rocket science.

vmsoftware:
I was trying this branch before merge and I can say it's really nice implementation. Great work! @javivelasco

However I have some small questions/suggestions:

1. What's about responsive table?  Maybe easy wrap the Table component into a div(for example),  set horizontal scroll and define columns widths 

2. What's the best way to handle múltiple sorted columns? 

3.  What do you think about a HOC called DATATABLE with  págination,  rows per page and header actions like MD spec? 

4. What's happend with the old table implementation?  I think it is very usefull too when we only need to show simple data.

javivelasco:
Thank you @vmsoftware!! Finally I merged because there is nothing you cannot do with the new implementation although I agree the old one is useful for some cases.

1. I totally missed it. It's a good opportunity to do it so I'm going to add it. I'm going to implement it or if you want to PR it would be great. 
2. Not sure. I'd need to think about it but since the source it's outside the component and you control sorting it's out of the scope. Probably a solution could be to sort by one criteria as in the example and identify repeated elements. Then you can sort those groups by the second criteria and so on.
3. I think it could be cool and useful. I don't have too much time to keep digging on this because there are other priorities but rather than an HOC I'd follow the same implementation of a wrapper component. It could be interesting. Since the `Table` structure is fixed and predictable, you can actually navigate the components from a wrapper to figure out children. Then you can render only what corresponds to the pagination parameters. Maybe it's too obscure but I can't think on a better approach keeping the same customization level.
4. Removed. The editing is definitively not a good idea because it breaks the guidelines. Generating data is really easy just by mapping rows and cells. It's a breaking change but with an straightforward migration if you're not editing anything.

ulrik59:
@javivelasco I would like to switch from Material-UI to React-Toolbox. I think to begin with `2.0.0-beta.0` is better since I will avoid migration but I am very interested in this new table implementation. I need a responsive data table with sorting and select features in Material Design. Is it possible to create a `2.0.0-beta.1` tag with this feature included ? Or should I directely use the `dev` or `cssnext` branches ?

javivelasco:
Of course, I'm doing it tomorrow because I want to add a new Layout too, keep tunned!

omaressameldin:
I was wondering where can i find an ecample for using the new table implementation?

NarasimhaReddyY:
@javivelasco

NarasimhaReddyY:
@javivelasco  Hey, You have mentioned like ```Now you can do easily stuff that was pretty difficult like having a sorted table, pagination, etc.```  but i couldn't see anything related to pagination that how we can do it easily?

javivelasco:
@NarasimhaReddyY Since the Table is composed from it's definition instead of generated internally you can do pretty much anything you want. The thing is that responsibility of sorting and paginating, etc, is now user responsibility. There are no examples because I think it should be trivial to do

PR #887, opened 2016-10-21
theme.required type should be string and not bool

juliobetta:


javivelasco:
Already fixed, I saw the other PR first :( Thanks anyway!!

PR #888, opened 2016-10-21
Build CalendarDay to reflect changes made in PR 871

rrott:
Built the latest dev branch to have the changes made in https://github.com/react-toolbox/react-toolbox/pull/871 be presented in the lib.

https://github.com/react-toolbox/react-toolbox/commit/5cea4e3a1bc2307ebd2be37ffa3ace85de84772a#commitcomment-19518131

### DIFF
--- a/lib/autocomplete/index.d.ts
+++ b/lib/autocomplete/index.d.ts
@@ -81,7 +81,7 @@ interface AutocompleteProps extends ReactToolbox.Props {
    * If true, the list of suggestions will not be filtered when a value is selected.
    * @default false
    */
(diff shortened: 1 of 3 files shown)


wtgtybhertgeghgtwtg:
Why is there even a `lib` folder, anyway?

javivelasco:
It's there just to allow people to clone the dev branch and use the same imports there. Do you think we can remove it??

wtgtybhertgeghgtwtg:
It might be a problem if they're pulling the package directly from GitHub, but that's about all I can think of.

PR #889, opened 2016-10-22
Fix Switch white-space bug

HaNdTriX:
### Bug

Text that doesn't fit in the viewport is currently hidden.

### Screenshot

bildschirmfoto 2016-10-22 um 14 44 15


javivelasco:
Thanks!

PR #893, opened 2016-10-23
Pass currently active suggestion to onBlur handler

seb0zz:
I currently have a use case where the current active selection is very useful for my onBlur handler. Don't know if you maybe see some constraints that speak against merging this. Otherwise it would be nice to have it in the main repository :-)


javivelasco:
Makes sense. Merged!

PR #949, opened 2016-11-08
Added enabled and disabled date properties to date picker.

dhamilton91:
Hi,

I added a property for enabled and disabled dates to the date picker component which both take an array of date objects.
Enabled dates will only enable all the dates in the array.
Disabled dates will only disable all the dates in the array.

What do you think?
This is useful for a project I work on as it will allow us to use two date pickers for a date range which only have dates enabled that we have data for.

David


javivelasco:
Great work! Just refactored a bit the code, apart from that it's very useful, thanks!

PR #968, opened 2016-11-16
added extension topic at readme.md

MaximKomlev:
added extension topic at readme for components which extend functionality of the library


javivelasco:
Thanks!!

PR #975, opened 2016-11-17
Implementing tab pagination per material design specifications.

therealparmesh:
Our company has many users who frequently rely on views with a lot of tabs or with tab labels that are long.

The current behavior for tab labels is for them to horizontally overflow their container if there are too many.

This pull request aims to resolve this issue by implementing tab pagination. For more information on why I went this route rather than just changing `flex-wrap` or `overflow`, please see the tabs section of the material design specifications (https://material.google.com/components/tabs.html).

Please let me know your thoughts on this.

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -33,7 +33,8 @@ const factory = (Tab, TabContent) => {
     };
 
     state = {
-      pointer: {}
+      pointer: {},
+      arrows: {}
     };
 
     componentDidMount () {
@@ -50,6 +51,7 @@ const factory = (Tab, TabContent) => {
       window.removeEventListener('resize', this.handleResize);
       clearTimeout(this.resizeTimeout);
       clearTimeout(this.pointerTimeout);
+      clearTimeout(this.arrowsTimeout);
     }
 
     handleHeaderClick = (event) => {
@@ -66,6 +68,7 @@ const factory = (Tab, TabContent) => {
 
     handleResizeEnd = () => {
       this.updatePointer(this.props.index);
+      this.updateArrows();
     };
 
     parseChildren () {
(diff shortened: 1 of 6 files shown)


javivelasco:
This is a great idea! I've been trying it out and I'd implement it slightly differently. While using it I noticed that moving back and forward moves the arrow with the same effect as when you change the tab. This can be prevented if the scroll container wraps the whole element including the pointer and it is absolute positioned from the parent. Also it would be more performant since we don't need to repaint the pointer when we move along the tabs.

What do you think?

therealparmesh:
Thank you for the feedback! I think I agree. It's a much smoother experience. I went ahead and made the changes.

therealparmesh:
Should be good to go!

javivelasco:
Great, thanks! I've made some changes like injecting the FontIcon, not render the arrows if they are not displayed and removing a couple of timeouts that I think are no longer needed.

Great work!

therealparmesh:
Awesome! Thank you for taking this in. Did not see that the `FontIcon` was exactly what I needed.

components/tabs/Tabs.js
```diff
@@ -147,10 +180,18 @@ const factory = (Tab, TabContent) => {
...
+          <div className={theme.navigationContainer}>
+            <div className={classnames(theme.arrowContainer, { [theme.invisible]: !(this.state.arrows.left || this.state.arrows.right) })} onClick={this.scrollNavigationLeft}>
+              <span className={classnames('material-icons', theme.arrow, { [theme.invisible]: !this.state.arrows.left })}>keyboard_arrow_left</span>
+            </div>
+            <nav className={theme.navigation} ref='navigation'>
+              {this.renderHeaders(headers)}
+              <span className={theme.pointer} style={this.state.pointer} />
+            </nav>
+            <div className={classnames(theme.arrowContainer, { [theme.invisible]: !(this.state.arrows.left || this.state.arrows.right) })} onClick={this.scrollNavigationRight}>
+              <span className={classnames('material-icons', theme.arrow, { [theme.invisible]: !this.state.arrows.right })}>keyboard_arrow_right</span>
```
kidkkr: why doese it shows shows `keyboard_arrow_right`?

rubenmoya: It's a material designs icon: https://material.io/icons/#ic_keyboard_arrow_right

kidkkr: Ah! Thank you. I saw the raw string sometimes when resizing `<Tabs />`

PR #980, opened 2016-11-18
Add "tooltipShowOnClick" to Tooltip component

cdroulers:
I couldn't figure out an easy way to do it without a PR in my own components.

I wanted this so I could still show a tooltip on mobile with the click / touch event. Otherwise tooltips are useless on mobile!

And it's smaller / easier than a full-on popup.


javivelasco:
Cool! Thanks :D

Issue #979, opened 2016-11-18
React update to 15.4.0


coljung:
So i've been working on a project for the last few weeks and everything is working fine.

today i tried to install said project somewhere else and came with all sorts of dependency errors regarding react@15.4.0

basically this was my package.json:

```javascript
"react": "^15.0.0",
"react-addons-update": "^15.3.0",
"react-dom": "^15.0.0",
....
"react-toolbox": "^1.1.1",
```

In my new location, it installed react-toolbox 1.2.5 and i got these messages:

```
npm ERR! peer dep missing: react@~0.14 || ~15.3.0, required by react-toolbox@1.2.5
npm ERR! peer dep missing: react-addons-css-transition-group@~0.14.0 || ~15.3.0, required by react-toolbox@1.2.5
npm ERR! peer dep missing: react-dom@~0.14.0 || ~15.3.0, required by react-toolbox@1.2.5
```

I tried then to manually install each one of them to a lower version, i couldnt get it to work in any way.

I tried something like:  npm i react@15.3.2 --save , and i got all these messages:
```
├── UNMET PEER DEPENDENCY react@15.3.2
├── UNMET PEER DEPENDENCY react-addons-css-transition-group@15.4.0
└── UNMET PEER DEPENDENCY react-dom@15.4.0

npm WARN react-addons-create-fragment@15.4.0 requires a peer of react@^15.4.0 but none was installed.
npm WARN react-addons-css-transition-group@15.4.0 requires a peer of react@^15.4.0 but none was installed.
npm WARN react-addons-update@15.4.0 requires a peer of react@^15.4.0 but none was installed.
npm WARN react-dom@15.4.0 requires a peer of react@^15.4.0 but none was installed.
npm WARN react-toolbox@1.2.5 requires a peer of react-addons-css-transition-group@~0.14.0 || ~15.3.0 but none was installed.
npm WARN react-toolbox@1.2.5 requires a peer of react-dom@~0.14.0 || ~15.3.0 but none was installed.
npm ERR! code 1
```

In any case, i spent a couple of hours this morning trying to figure out what to do, and i ended up modifying my current package.json so something like this:

```javascript
"react": "~15.3.0",
    "react-dom": "~15.3.0",
    "react-addons-create-fragment": "~15.3.0",
    "react-addons-css-transition-group": "~15.3.0",
    "react-addons-update": "~15.3.0",
```

Just wondering if anyone had something similar happen to them and how did they fix it ?

[...]

Issue #970, opened 2016-11-16
Bump `react` peerDependency.


wtgtybhertgeghgtwtg:
`react@15.4.0` is out, but `react-toolbox` [does not accept it as a peerDependency](https://github.com/react-toolbox/react-toolbox/blob/dev/package.json#L120).

ToniIltanen:
I just noticed the same thing.

eskimoblood:
Is there any reason to lock it to a minor version at all. I think we can trust the react team to follow semver, so why not just lock it to the major version.

larsts:
Same goes for react-addons-css-transition-group

PR #981, opened 2016-11-18
Restrict dependencies to major versions only.

borela:
All dependencies are following semver, so we should assume that only bumps into major versions would break this project.

This pull request relax the dependencies a bit without allowing breaking changes.


borela:
Fixes #979 and #970.

kockok:
Hello, is the problem solved? 
I still have these warnings.
Or should I just ignore them?


> npm WARN react-toolbox@2.0.0-beta.8 requires a peer of react@^0.14 || ~15.4.0 but none was installed.
npm WARN react-toolbox@2.0.0-beta.8 requires a peer of react-addons-css-transition-group@^0.14.0 || ~15.4.0 but none was installed.
npm WARN react-toolbox@2.0.0-beta.8 requires a peer of react-dom@^0.14.0 || ~15.4.0 but none was installed.

javivelasco:
They are peer dependencies so you need to add them manually to your project. It is assumed that if you use React Toolbox you'll have React installed :)

chiester:
Is it the case that react-toolbox itself uses React 15.4.1, but requires projects that use it be on 15.4.0 or earlier?

javivelasco:
Yeah, as stated in the `peerDependencies`. Anyway I think other versions would make it work but you will get a npm/yarn warning when installing

kockok:
The warning dismissed after changing `^15.4` to `~15.4.0`. Haha

PR #990, opened 2016-11-22
Fix Autocomplete mouse selection.

### LINKED ISSUE #986: Autocomplete onChange returns event, not value.
My Autocomplete component has this onChange prop `onChange={(value) => {console.log(value)}}`.  Before, it returned a value, now it returned an event object but the target property is null. The input text component is working, only the auto complete is not.

borela:
Fixes #986 which probably happens on the css-next branch too.


javivelasco:
Thanks! :D

PR #991, opened 2016-11-22
Adds onQueryChange callback property to Autocomplete

motevets:
The onQueryChange callback is called when the value of the query changes in Autocomplete.
It is called with the new query value.


javivelasco:
Thanks!

PR #992, opened 2016-11-22
Auto build on install

borela:
Builds the library after installing it.

NPM Tested:
- [x] Windows
- [x] Mac
- [x] Linux

To do:
- [x] Download only the necessary dependencies to build.

Known issues:
* `npm update` clones the repo even if there are no changes to it which would cause a rebuild (this is NPM's fault).

To test it:

`npm i borela/react-toolbox#autobuild`

### DIFF
--- a/package.build.json
+++ b/package.build.json
@@ -1,6 +1,7 @@
 {
   "dependencies": {
     "autoprefixer": "^6.5.1",
+    "babel-cli": "^6.18.0",
     "babel-core": "^6.18.2",
     "babel-loader": "^6.2.4",
     "babel-plugin-transform-class-properties": "^6.19.0",


thomasthiebaud:
It seems to work well on Mac (version 10.12.1)

.babelrc
```diff
@@ -1,19 +1,14 @@
...
-            "locals": ["module"]
-          }, {
-            "transform": "react-transform-catch-errors",
-            "imports": ["react", "redbox-react"]
-          }]
-        }]
-      ]
-    }
-  }
+  "presets": [
```
wtgtybhertgeghgtwtg: Couldn't you just use `babel-presets-latest` instead of `es2015`/`es2016`/`es2017`?

borela:
@wtgtybhertgeghgtwtg yep it does, I changed that.

package.json
```diff
@@ -47,11 +48,13 @@
     "babel-core": "^6.18.2",
     "babel-eslint": "^7.1.1",
     "babel-loader": "^6.2.4",
-    "babel-plugin-react-transform": "^2.0.2",
+    "babel-plugin-transform-class-properties": "^6.19.0",
+    "babel-plugin-transform-decorators-legacy": "^1.3.4",
```
javivelasco: Are we using this?

borela: Decorators I didn't saw, I'll remove it but class properties are being used.

borela:
The main issue right now is that it'll build on each install, not only the git ones.

package.json
javivelasco: Yeah, I meant decorators. I think we are only using Object rest spread, class properties and maybe export extensions.

javivelasco:
Ups... Not cool, maybe it's too much friction?

borela:
If the postinstall comand targets the lib directory it won't execute as intended, but if you update the package without deleting the lib dir, there's no way to tell if it's being installed from npm vs git.

borela:
I think if it deletes the lib dir on preinstall it should work as intended, I'll check that.

thomasthiebaud:
It seems to work on Linux also (Ubuntu 16.10)

thomasthiebaud:
If I run `npm ls react`, I get 

```
thinxmateui@0.1.0 /home/thomasthiebaud/Combain/mate-v2-web
├── react@15.4.1 
└─┬ react-toolbox@1.2.5 (git://github.com/borela/react-toolbox.git#37523f03c40d7de9fa5d0482d70fae968ffdd9ee)
  └── react@15.4.1  extraneous
```

As you can see there is two versions of react, which causes this issue

> Only a ReactOwner can have refs. You might be adding a ref to a component that was not created inside a component's `render` method, or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).

borela:
@thomasthiebaud postinstall wasn't deleting the temporary node_modules it created under the lib's dir, I added a command to remove it and that problem disappear. 

@javivelasco Added a test lib dir to the repo and it worked, so it shouldn't tried to build on any release containing a lib dir.

If you install the lib from github and try to update it, it'll rebuild even if there's no new commits because npm clone without checking if there are any changes to the repo.

wtgtybhertgeghgtwtg:
`postintall-build` seems to [download all `devDependencies`](https://github.com/exogen/postinstall-build/issues/11).  So that would include `mocha`, `karma`, `express`, and several other large things that would not be used.

borela:
@wtgtybhertgeghgtwtg It does but it creates a temporary node_modules under the lib's dir: `your project/node_modules/react-toolbox/node_modules` which is deleted after the build process.

wtgtybhertgeghgtwtg:
But it still downloads them.

borela:
Yeah, I would be nice if we could specify the "build only" dependencies in the package.json.

wtgtybhertgeghgtwtg:
I would say that it's a requirement.  Why should the user have to download a linter, a server, or a testing framework just to delete it?

borela:
I do agree and I need a solution for that, maybe we could send a pull request to postinstall to e.g read a build.json to install only the needed packages or write a specific script for this project to build on demand.

Another problem I found is that sometimes yarn chokes with `cpx`.

borela:
I created a [pull request to the postinstall ](https://github.com/exogen/postinstall-build/pull/12) and with those changes it installed only files listed in [package.build.json](https://github.com/borela/react-toolbox/blob/autobuild/package.build.json), It worked at least on windows.

It needs testing but it should be faster as instead of pruning the directory, it just deletes it with `--cleanup` flag added to the postinstall command.

javivelasco:
I'm re-reviewing PRs and this cannot be included in the re-organization I'm planning on a different branch. The reason is that react toolbox is going to be distributed as many different packages hosted in a monorepo using [lernajs](https://lernajs.io/). For this, the scripts for building are no longer the same and also it changed a bit with gulp. Anyway I guess if you want to use the repository version you can just go and run the build yourself.

As soon as I published this reorganization (which should be happening during this next week), I'd love you get your feedback about it. I guess it doesn't make any sense to a monorepo anyway. Thank you so much for dedicating time, I always appreciate it

PR #994, opened 2016-11-24
Pass event to selection.

### LINKED ISSUE #993: Autocomplete vs NODE_ENV=development
Thx for great project!

When bundling my app with NODE_ENV=development, Autocomplete fails onChange behavior
It's because in Autocomplete:232 (search by text "select = (event, target)") event becomes Proxy and target becomes event. In that case "const newValue" resolving logic returns target.

Am I correct? Or this is my fault (incorrect usage of Autocomplete?)

borela:
The event must be passed as a parameter.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -96,8 +96,8 @@ const factory = (Chip, Input) => {
      }
    };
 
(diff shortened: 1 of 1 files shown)


javivelasco:
Thanks!

PR #666, opened 2016-07-24
Migrate styles to PostCSS

javivelasco:
This PR brings a complete migration of react-toolbox to [post-css](https://github.com/postcss/postcss) along with changes that are side effects of this migration. The theming philosophy behind this migration is basically the same as before. Each component has a configuration file where some variables are defined and used at time of generating styles. Those variables can be overridden to customize appearance. We are using:
- [postcss-import](https://github.com/postcss/postcss-import) allows to separate CSS in multiple files so we can do stuff like importing variables from one component into another.
- [postcss-mixins](https://github.com/postcss/postcss-mixins) allows to define and use mixins. They are not very used in react-toolbox but I've found them useful in the `Layout` component since some properties need to be repeated for different screen sizes.
- [postcss-next](http://cssnext.io/) the most important one. Includes multiple plugins to define custom variables, to use `calc`, etc.
- [postcss-each](https://github.com/outpunk/postcss-each) used in the `Layout` component to generate a bunch of classes that define sizes of the drawer element, for example.

With this implementation we don't need to [make tricks](https://github.com/react-toolbox/react-toolbox/blob/dev/docs/webpack.config.development.js#L52-L54) to prepend variables SASS files to react-toolbox stylesheets. We just need a [variables Javascript file](https://github.com/react-toolbox/react-toolbox/blob/css-next/docs/toolbox-variables.js) defining theming variables we want to override and [pass it to the custom properties plugin configuration](https://github.com/react-toolbox/react-toolbox/blob/css-next/docs/webpack.config.development.js#L65-L71). In case you are using sass variables from react-toolbox now, with the new version you'd need to change to postcss. Apart from that, if basically works the same way.

Of course this is a **huge** change because it implies a technology change and I think it should be released under `2.0.0`. The amazing news is that we can do awesome stuff with this change that I'd like to include in this PR before merging:
- [ ] Update MD documenting the usage of react-toolbox with PostCSS.
- [ ] Review and document the custom properties used for each component.
- [ ] More detailed documentation with specific info for Webpack and Theming.
- [ ] Redesign the docs.
- [ ] Include theming live customization in the docs playground.
- [ ] Create a live theme generator in the documentation.

Please leave your feedback in this PR!

PR #1000, opened 2016-11-27
Split component theme's style from imports

razor-x:
This moves all non-import statements in each component's
theme.css into style.css (which is then imported into theme.css).

This change is transparent to users and is organizational only.

This allows one to override the variables used in a component theme since one can now import the styles without the variables.

I only changed files called `theme.css`: the animation styles still have the same limitation and would need to be considered separately.

See https://github.com/react-toolbox/react-toolbox/pull/666#issuecomment-259304780.


razor-x:
@javivelasco Updated with the latest changes on css-next. Let me know if this is an approach you want to adopt.

javivelasco:
Sorry for the late response @razor-x 

I actually think this makes sense, but if I'm not wrong, with custom properties you can actually add your custom variables at the end of the `theme.css` file. In this case, these changes are not needed and it would be cleaner to keep it without split.

If this is not possible I'll happy to merge but I think this is the how the spec was written.
Note that it's not the same for variable definitions because they might be needed in a dependent component.

razor-x:
I wasn't aware of that behaviour. I can play around with it a bit more to see if this is still needed.

javivelasco:
Cool, let me know!

razor-x:
Ok, that's really :cool:. Works just as you said and properties are still computed from other vars correctly, e.g., make a small button with

```css
@import 'react-toolbox/lib/button/theme.css';

:root {
  --unit: 10px;
}
```

javivelasco:
Cool! 💃

mikkom:
@razor-x Can you elaborate on how you got this working? For example, what would the SmallButton component look like in this case?

mikkom:
Or can you @javivelasco maybe point me to instructions on how to create multiple customized versions of React Toolbox components with v2? Thanks in advance!

mikkom:
So far the best I've managed is the following:

`CustomizedButton.js`

```javascript
const theme = {
    RTButton: require('./CustomizedButton.css')
};

const CustomizedButton = props => (
    <ThemeProvider theme={theme}>
        <Button {...props} />
    </ThemeProvider>
);
```

`CustomizedButton.css`

```css
@import './theme.css';

.button {
    width: 300px;
}

.primary:not([disabled]) {
    &.raised,
    &.floating {
        background: wheat;
    }
}
```

and then copying react-toolbox/lib/button/theme.css to same directory and helpers.css (which theme.css refers) to the parent directory. If I use `@import 'react-toolbox/lib/button/theme.css';` then I get a failure from theme.css trying to refer to '../helpers.css'. Any hints on how to avoid copying of the styles would be really helpful.

Also, is using the ThemeProvider here the best approach? Could I easily achieve same effect by passing the theme somehow via props to Button component?

javivelasco:
Actually is quite easy once you get how styles are propagating. The thing is that it also depends on how are you requiring the component (at least until now, let's see if there is a better approach for future). Assuming you are using CSS Modules:

1. You can import the component with styles and create your own custom CSS Module with the overrides. In this case the selector priority should be higher than the original and you must pass the theme as you did in your previous snippet. In that scenario the component gets merged both the original and your custom CSS Module classnames. 
2. You can import the component raw and define a custom CSS module that imports **the original css file**. Then you can add custom variables at the bottom of the file which is the subject of this issue. This may be inconvenient if you are not changing a lot because different versions with different CSS modules would be generating the same styles with different classnames as many times as modules you have. For example ButtonA.css and ButtonB.css imports Button.css from react-toolbox. Styles from original Button.css would be bundled twice but processed with different variables. I prefer 1 over this.
3. Similar to 1 but you set the default styles in the context with `ThemeProvider`. This is the way to go with react-toolbox-themr and works the same as the first one. The only thing inconvenient is that the original CSS is not auto-bundled but required imperatively from the App root.

In any case you need to define a custom module with the overrides or every classname in the component. If you want the overrides you need to pass the originals by some way, either importing with the css or using ThemeProvider. If you want to pass a full theme you can import original CSS from your custom CSS but in this case make sure you import the component raw and you don't keep more than two or three custom components duplicating the original styles.

Hope it's understandable!!

PR #1004, opened 2016-11-30
Dropdown auto fix

makieno:
Dropdown's property `auto` isn't working (anymore), and always opens down even if the Dropdown is positioned in the lower half of the screen.


drummerhu:
Also fixes that dropdown does not open in firefox (event is not defined)

javivelasco:
Thanks! Sorry for the miss

PR #1014, opened 2016-12-01
1.3.1 autocomplete improvements

craigcartmell:
Couple of autocomplete changes here:

* Exposes onQueryChange val to parent
* Allows and handles value to be passed as object of key/val pairs (similar to source) as well as the current string and array options.

Finally, it adds a showSelectedWhenNotInSource bool prop. 

Take the current scenario:

You have a cities table which contains 60,000+ records so you don't want to fetch and render them in the autocomplete at once. Instead you fetch the first 15 and onQueryChange, fetch the ones which match the query and merge this into the autocomplete source.

Currently this means any previously selected values which don't exist in the source currently won't show in the chips.

Passing showSelectedWhenNotInSource fixes this.


javivelasco:
Cool, thanks for this PR!

Just an small comment: There is a complete rewrite of this component coming. Some of the issues you are covering with this PR will be resolved just by reducing the encapsulation. For example, next version will no longer support a source, that the rendering of selected items will be delegated outside the component.

Meanwhile... thanks! :)

javivelasco:
One more thing, may you update the Readme with the new props?

craigcartmell:
Hey @javivelasco ,

Thanks for the feedback, I'll update the readme now.

Are the changes you mentioned upcoming in react-toolbox 2.0?

craigcartmell:
Made the changes to the readme @javivelasco 👍

javivelasco:
Thanks @craigcartmell, sorry to bother but there are conflicts now, probably because of another PR I merged yesterday. These autocomplete changes will be probably out of 2.0 but for the next iteration :)

craigcartmell:
@javivelasco Whoops, conflicts fixed.

PR #1023, opened 2016-12-02
Tooltip fixes

Zequez:
- When creating a tooltipped component sometimes you might want to add a tooltip into a simple native component, which throws an error when it tries to pass the `theme` prop. Now creating a tooltip like `const TooltippedDiv = Tooltip('div')` works as expected.
- The tooltip now uses the DOM node of `currentTarget` instead of `target` when calculating the position. This makes the tooltip centered in the correct place when adding it to a component that has multiple children in different positions.


javivelasco:
Beautiful! Thank you so much!

PR #1024, opened 2016-12-02
fixed of appearing dialog at IE

MaximKomlev:
Fixed of appearing dialog at IE (style note defined exception)
Added autoHeight properties to dialog so dialog can be expanded on whole content and scroll bar will appear on overlay, so the option allows to make dialog being more flexible to content and in case of huge content (let say document preview) on mobile screen. 
Sample is added to spec (dialog shows changelog).

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -162,6 +162,7 @@
       "location": "start"
     }],
     "no-with": [2],
+    "object-shorthand": [2],
     "one-var": [0],
     "operator-assignment": [0, "always"],
     "operator-linebreak": [2, "before"],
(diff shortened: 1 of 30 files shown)


javivelasco:
I appreciate this a lot but the PR is including changes for the `Autocomplete`. Also, the new implementation will change a bit and there are tons of conflicts. If you've found this very useful for your use case I'd appreciate if you can send an updated PR including just the changes for the Dialog and related (Overlay) components.

Thank you for this!

MaximKomlev:
Ok, I see, I will adapt the changes for new implementation.

Best Regards, Max.

Issue #920, opened 2016-11-01
unable to style input child of autocomplete


t22james:
this seems similar to this previous ticket:
https://github.com/react-toolbox/react-toolbox/issues/674
In v1.1.2 I am able to pass theme={theme} to DatePicker, and the nested Input (including labels, errors etc) also receives all of the themed classes (even though i can't see it explicitly passing it down as a prop). I expected the same thing to occur with Autocomplete, but it does not. The nested Input (including labels, errors etc) doesn't receive the theme despite same syntax/prop. Could both of these elements (and any others that nest an Input) perhaps have an inputTheme prop added, to explicitly pass desired full theming abilities down into the input (including labels, errors etc)?

t22james:
https://github.com/react-toolbox/react-toolbox/issues/877
potentially duplicate, but that ticket only specifies 'label'. I'd like entire Input (i.e. with all children) exposed to a theme, so i can normalise with my standard Input fields.

rubenmoya:
Yep, probably adding `className={classnames(this.props.theme.input, {[inputClassName]: inputClassName })}` to the autocomplete Input (same as DatePicker) will solve it, since you could use `.input {}` in your theme.

javivelasco:
Fixed. You can write classNames to theme the `Input` component namespaced under `input` keyword. Check the readme in `Autocomplete`!

t22james:
@javivelasco ok finally uncovered what you mean by reading the readme file in the 1.2.5 Autocomplete folder itself rather than online. This still doesn't seem to work as well as the method above, but understand you must have reasons for doing so, to do with namespacing etc. 

There is further confusion currently in that the current version, 1.2.5 on npm seems to contain /lib files which don't match the /components, so anyone running react-toolbox from npm will not be getting that functionality by default.

Finally, if this is going to be the chosen method for handling nested styles, then Dropdown and Datepicker should also receive the same support?

javivelasco:
Sor sure. There are components that need to be updated with this approach too. It's relatively new and not implemented in react toolbox but it's very handy. I'm going to run the build and deploy during the next days to keep everything updated!

stereokai:
@javivelasco @t22james I was having a hard time coping with theming in react-toolbox and I thought I could add some information because the explanation of this feature is quite unclear:

> ### Theme
>
[...]

Issue #877, opened 2016-10-18
Autocomplete ignores label theme


igorbarbashin:
I'm trying to customize label on Autocomplete field with CSS Modules theme.

In my css file I have:

``` css
.label {
  color: pink;
}
```

And I pass it like:

``` jsx
import theme from './theme.css';
<Autocomplete
  theme={theme}
/>
```

But label color doesn't change.

I believe in order to fix it we need to pass theme to Input component of the Autocomplete (because Input renders the actual label):
https://github.com/react-toolbox/react-toolbox/blob/dev/components/autocomplete/Autocomplete.js#L300

Something like:

``` jsx
<Input
  theme={{
    label: theme.label
  }}
  ...
/>
```

PR #1027, opened 2016-12-05
Enable theming of Input component inside Dropdown component.

t22james:
Uses the same method recently implemented on the Autocomplete component.
Related to #877 and #920 and commit d0a9032fe7c616fcce78c50865d134c87c1b9746


t22james:
Realise you have a rather major release pending, but would love if you could fire through this quick fix through on a minor. I've just followed your recent lead on the Autocomplete Input theming, seems to be working the same on Dropdown, so why not start using? :)

javivelasco:
Sure, I'm releasing a patch in a couple of days!

PR #1033, opened 2016-12-07
Autocomplete isValueAnObject cleanup

craigcartmell:
Cleans up the isValueAnObject logic. I thought `this.props.value` was being converted when I initially wrote this.

I was finding sometimes `this.state.isValueAnObject` wasn't being set correctly intermittently so this fixes that too.


javivelasco:
Thanks!

PR #1035, opened 2016-12-07
Add ability to sort a data table by clicking a table heading

tuckerconnelly:
Adds the ability to sort columns by clicking the table header. Accepts an `onClick` in the model, which will make the column clickable. Then, you can pass `props.sortColumn` and `props.sortDirection` into the `<Table />` to get the up/down icon to display.


javivelasco:
Thank you @tuckerconnelly I have been keeping this unmerges because the Table is already deprecated in favour of a more flexible version. We are keeping the 1.x in the master branch so if you think it's worthy feel free to open a PR there. Dev holds now the 2.0 beta with some rewritten components

PR #1041, opened 2016-12-11
Slider debounced

landabaso:
Give the possibility to debounce touchmove and mousemove events using lodash.debounce. Useful when using the Slider with not-so-new mobile devices.

### DIFF
--- a/components/slider/Slider.js
+++ b/components/slider/Slider.js
@@ -7,11 +7,15 @@ import events from '../utils/events.js';
 import utils from '../utils/utils.js';
 import InjectProgressBar from '../progress_bar/ProgressBar.js';
 import InjectInput from '../input/Input.js';
+import debounce from 'lodash.debounce';
 
 const factory = (ProgressBar, Input) => {
   class Slider extends Component {
     static propTypes = {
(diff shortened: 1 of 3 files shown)


landabaso:
While this request is not accepted (or in case it never makes it), note that it is possible to override the touchmove/mouse move methods:
```
import debounce from 'lodash.debounce';
import {Slider as internalSlider} from 'react-toolbox/lib/slider/Slider';
import Slider from 'react-toolbox/lib/slider/Slider'; //this is the themed one you'll probably use.

const oldComponentDidMount = internalSlider.prototype.componentDidMount;
internalSlider.prototype.componentDidMount = function() {
  this.handleMouseMove = debounce(this.handleMouseMove, 32, {maxWait: 200});
  oldComponentDidMount.bind(this)();
};
```

It's hacky and not recommended if you don't know what you are doing. Using this hack you won't need to patch the original npm package.

In my experience, debouncing these calls[ is a must](https://css-tricks.com/debouncing-throttling-explained-examples/), specially for old devices.

javivelasco:
Sorry I missed to merge this. Can you solve conflicts to get it merged?

PR #1042, opened 2016-12-12
relaxing propTypes of Input & Autocomplete

masakij:
This PR relaxes following propTypes of Input & Autocomplete.

* error: string -> node
* hint: string -> node
* label: string -> node

This is intended for use in following use cases.

* awesome styled label
* interactive label
* some translating or formatting component (like react-intl)

For example
```jsx
const error = <span>ERROR!! <a href="/help">?</a><span>;
<Autocomplete error={error} />;
const label = <span style={awesomeStyle}>label</span>;
<Autocomplete label={label} />;
const hint = <FormattedMessage />;
<Autocomplete hint={hint} />;
```


masakij:
OK, I'll change the propTypes to it.

(IMO, It's a bit redundant because PropTypes.node includes PropTypes.string. https://facebook.github.io/react/docs/react-api.html#react.proptypes.node )

javivelasco:
Thanks!

PR #1056, opened 2016-12-15
Dropdown: add generic overflow, segmented and editable options

Ehres:
This PR add last options of Dropdown component described in [material guideline](https://material.google.com/components/buttons.html#buttons-dropdown-buttons)

- Generic overflow dropdown button
- Segmented dropdown button
- Editable segmented dropdown button


javivelasco:
Thanks! May you fix conficts and complete the docs for these new props? Also recall the Dropdown is going to be completely re-written for the 2.0 so this would be kept in 1.x meanwhile

javivelasco:
I'm closing since it looks like there is no activity on this. Since it's going to be rewritten I'd use it as a reference for this future version. Anyway if you want to update it and solve conflicts I'll be happy to reopen and merge while the new version is finished.

PR #1062, opened 2016-12-16
Remove commons

javivelasco:
This PR removes `commons.css` so all components should work with the only css required by them. Of course this would not break your styles anymore. Please @okonet try it out if you want and we can merge it.

I need it for something I'm preparing... :)


okonet:
LGTM!

PR #1066, opened 2016-12-18
Declare immutability-helper as direct dependency

just-boris:
As a library user, I should not care about some library implementation details, e.g. that it uses `immutability-helper`.

As well as `classnames`, which is already defined in `dependencies`, all other libraries, except `react-*` should be also installed automatically, when you are running `npm i react-toolbox`


javivelasco:
Thanks, it's already done in postcss version

PR #1067, opened 2016-12-19
Auto-generated Update to Changelog

Klathmon:
I used a package called [`github-changes`](https://github.com/lalitkapoor/github-changes) which i've used with good results in the past to auto-generate a changelog for everything from 1.1.2 to the current 1.3.3 release.

[Here is what it will look like on github](https://github.com/react-toolbox/react-toolbox/blob/4c817670cca27831c86e1e9f14026259b78de3af/CHANGELOG.md)

It's far from perfect, and it's in a different format than your past changelogs, but it does at least get everything caught up. They can also be copied verbatim to the release page on github and it will auto-link usernames and issue/PR numbers for you there as well.

If you start including the title of the issue that gets fixed in the commit message, it will really round it out to a nice looking changelog, as commits like "Fixed ####" would give more context without any significant work.

Just in case you want it, here is the exact command I used with my github token removed:

    github-changes -o react-toolbox -r react-toolbox --branch dev --issue-body --use-commit-body --token xxxxtokenxxxx 

I've had good success with using this in a `post-version` npm script, but with some tweaking (mainly to have it only generate the changelog for the previous tag, and some command-line work to prepend it to the current changelog file and not overwrite anything).


Kikobeats:
The changelog is auto generated in each release using [bumped plugin](https://github.com/bumped/bumped-changelog) for that.

In your changelog beta releases are not present, so what is the point for use a different tooling?

Klathmon:
I wasn't in a position to give a changelog for the 2.0 changes, so i didn't.

I just wanted a changelog for the 1.x changes (specifically the 1.1.2 to 1.3.3 versions) and since they didn't exist I made them myself using this tool. Decided to share it just in case it was useful to anyone.

Now that 2.0 is going to be the only supported branch moving forward, i'll just close this.

PR #1070, opened 2016-12-19
update typings to 1.3.3

katzoo:
This PR introduces few updates and improvements to TS typings.

First of all I exposed (added `export`s) to all interfaces as it doesn't hurt (non-class/component interfaces have appropriate suffix (`...Theme` or `...Props`)) and it helps a lot to extend or apply custom themes to components. Because of `react-css-themr` typings, any typed component that goes through it will be of `any` type, but it's easy to fix that with exposed interfaces like so:
```tsx
import { Button, ButtonProps } from 'react-toolbox/lib/button';
const MyThemedButton: React.ComponentClass<ButtonProps> = themr(BUTTON_ID, defaultTheme)(Button);
// enjoy TS features on <MyThemedButton /> :-)
```
In my opion exporting all interfaces also allows for easier fixes/extensibility in the future - by just merging declarations (when new properties are introduced) or using Partials in TS 2.1 (if properties are deprecated). Both are much better options than copy-pasting and fixing typings locally.

I also updated typings to correspond with current react-toolbox version. I've tried to follow the code (propTypes + passing/extending props), so most of updated types should work just fine. It's possible that I've missed something, but TS shouldn't cry about missing properties. As for actual code I'm not 100% sure about my typings for `List` and `Ripple`, but I can update/create a new PR if anything comes).

`form` and `hoc`(ActivableRenderer + Portal) directories were left with no typings intentionally. While HOCs are used internally by other components (and probably not a lot of TS usecases here), the `form` looks kinda abandoned (it's not listed on the docs page + even markdown is a bit off). Also I didn't bother to split bigger components like time picker into separate ones (it was like this before).

Additionaly I've found that some properties are missing from readmes, so added a few (will submit another PR), but some need to be rewritten from scratch (eg. List) (just in case I'm not really a markdown guy, so it will wait for somebody else :smile:).

### DIFF
--- a/components/app_bar/index.d.ts
+++ b/components/app_bar/index.d.ts
@@ -17,7 +17,7 @@ export interface AppBarTheme {
   /**
    * Used as the app bar title.
    */
-  title ?: string;
+  title?: string;
   /**
    * Added to the left icon app bar element.
    */
@@ -26,9 +26,13 @@ export interface AppBarTheme {
    * Added to the right icon app bar element.
    */
   rightIcon?: string;
+  /**
+   * Added to the root element when the app bar is hidden during scroll.
+   */
+  scrollHide?: string;
 }
 
-interface AppBarProps extends ReactToolbox.Props {
+export interface AppBarProps extends ReactToolbox.Props {
   /**
    * Children to pass through the component.
    */
@@ -50,7 +54,7 @@ interface AppBarProps extends ReactToolbox.Props {
   /**
    * If it exists it is used as the AppBar left icon
    */
-  leftIcon?: string;
+  leftIcon?: string | React.ReactNode;
   /**
    * Called when the left icon is clicked
    */
@@ -58,11 +62,16 @@ interface AppBarProps extends ReactToolbox.Props {
   /**
    * If it exists it is used as the AppBar right icon
    */
-  rightIcon?: string;
+  rightIcon?: string | React.ReactNode;
   /**
    * Called when the righticon is clicked
    */
   onRightIconClick?: Function;
+  /**
+   * Whether AppBar should be hidden during scroll.
+   * @default false
+   */
+  scrollHide?: boolean;
   /**
    * Classnames object defining the component style.
    */
--- a/components/autocomplete/index.d.ts
+++ b/components/autocomplete/index.d.ts
@@ -1,5 +1,6 @@
 import * as React from "react";
 import ReactToolbox from "../index";
+import { InputTheme } from "../input"
 
 export interface AutocompleteTheme {
   /**
@@ -18,10 +19,6 @@ export interface AutocompleteTheme {
    * Used to style the Input component.
    */
   input?: string;
-  /**
-   * Used for the label.
-   */
-  label?: string;
   /**
    * Used to style each suggestion.
    */
@@ -44,7 +41,12 @@ export interface AutocompleteTheme {
   values?: string;
 }
 
-interface AutocompleteProps extends ReactToolbox.Props {
+export interface AutocompleteProps extends ReactToolbox.Props {
+  /**
+   * Determines if user can create a new option with the current typed value.
+   * @default false
+   */
+  allowCreate?: boolean;
   /**
    * Determines the opening direction. It can be auto, up or down.
    * @default auto
@@ -52,6 +54,7 @@ interface AutocompleteProps extends ReactToolbox.Props {
   direction?: "auto" | "up" | "down";
   /**
    * If true, component will be disabled.
+   * @default false
    */
   disabled?: boolean;
   /**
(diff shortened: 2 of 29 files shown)


javivelasco:
Cool, thanks!

newash:
Excuse me if I got anything wrong here, but this change breaks existing code and I don't see any workarounds. I have compile errors like this using TS2.1:

`error TS2339: Property 'caption' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<ListItem> & { children?: ReactNode; } & ListItemPr...'.`

Not only for caption but for all the other attributes you removed from ListItemProps.

Even more, I fail to see how exactly this was supposed to work. I'm having this import:

```
import { List, ListItem, IconButton } from "react-toolbox";
```

Now the ListItem class is defined in the d.ts with ListItemProps only. ListItemProps itself is defined by knowing nothing about ListItemLayoutProps or ListItemContentProps. So how TypeScript is supposed to know about the attributes there?
More importantly: what was your idea for developers to have this working, what should I do?

Thanks!

newash:
Also with the `Ripple`:

In the top-level `index.d.ts` there is:

```
import { Ripple } from "react-toolbox/lib/ripple";
```

only the type definition file there does not have any "Ripple" definition, so it's another compile error:

`error TS2305: Module '"C:/DEV/visu-react-pilz/node_modules/react-toolbox/lib/ripple/index"' has no exported member 'Ripple'.`

(@javivelasco Maybe it's overkill, but some automated test would help so changes with compile error wouldn't end up in a release.)

panjiesw:
@newash I agree about including some tests for typescript definition. I'm working on another ts definitions update in #1114 so maybe I'll include it there. Also with fix to the issue you mentioned above.

For now, I think you can use declaration merging for that particular module. Create a `react-toolbox.d.ts` at the root of your project and declare module definition. I'm on mobile now, I'll post an example here when I'm home.

newash:
@panjiesw Thanks for the quick reply!

panjiesw:
@newash please try this temporary workaround and let me know if it isn't working

```typescript
// react-toolbox.d.ts
declare module 'react-toolbox/lib/list' {
	import * as React from 'react';
	import { ListProps } from 'react-toolbox/lib/list/index';
	import ReactToolbox from "react-toolbox/lib/index";

	export interface NewListProps extends ListProps {
		// Add here any other properties missing from original ListProps
		// e.g. caption
		caption?: string;
	}

	export class List extends React.Component<NewListProps, {}> { }

	// Add any other definition you need here.
	// Basically override the 'react-toolbox/lib/list' definition with your own
	// defined here.
}
```

newash:
Thanks for getting back! I've tried your workaround, but unfortunately TypeScript doesn't seem to merge the two module declarations. It's just taking this custom one, and forgets about the original file, then it complains that other definitions are missing from 'react-toolbox/lib/list'

```
ERROR in xxxxxxxxxxxxxxxxx\node_modules\react-toolbox\index.d.ts
error TS2305: Module ''react-toolbox/lib/list'' has no exported member 'ListCheckbox'.
error TS2305: Module ''react-toolbox/lib/list'' has no exported member 'ListItem'.
...
```

I was trying too google it what could have been the problem, but no success so far.
I guess I'll roll back to the previous version for the time being instead of the workaround.

Thanks for your kind help.

panjiesw:
Yes, that's the cons of this workaround, you need to include all you need in that custom definition file. Just copy them from the original. This make you able to customize the types outside of the `react-toolbox` package.

katzoo:
Hey @newash, sorry for causing some problems. It's totally my fault - I didn't double check these against tsc. As for `List` have rewritten definitions based on `propTypes`, but didn't really follow how they are passed around to sub/nested components. The list issue was that `caption` now lives on `ListItemLayoutProps` so these need to be merged for `ListItem` component.

#1114 Should fix these issues, but as @panjiesw suggested for temporary workaround you can overwrite broken definitions.
Simply create a new `.d.ts` file (eg. `react-toolbox.d.ts`) and make sure your `tsconfig.json` doesn't exclude it (I personally put these in `typings` directory).
```ts
declare module 'react-toolbox/lib/ripple' {
  import {
    default as Ripple,
    RippleProps,
    RippleTheme,
  } from 'react-toolbox/lib/ripple/index';

  export default Ripple;
  export { Ripple, RippleProps, RippleTheme };
}

declare module 'react-toolbox/lib/list' {
  import React from 'react';
  import ReactToolbox from 'react-toolbox/lib/index';
  import {
    ListCheckboxProps,
    ListDividerProps,
    ListItemActionProps,
    ListItemActionsProps,
    ListItemContentProps,
    ListItemLayoutProps,
    ListItemProps,
    ListItemTextProps,
    ListProps,
    ListSubHeaderProps,

    ListCheckboxTheme,
    ListDividerTheme,
    ListItemActionsTheme,
    ListItemActionTheme,
    ListItemContentTheme,
    ListItemLayoutTheme,
    ListItemTextTheme,
    ListItemTheme,
    ListSubHeaderTheme,
    ListTheme,
  } from 'react-toolbox/lib/list/index';


  export class List extends React.Component<ListProps, {}> { }
  export class ListCheckbox extends React.Component<ListCheckboxProps & ListItemContentProps, {}> { }
  export class ListDivider extends React.Component<ListDividerProps, {}> { }
  export class ListItem extends React.Component<ListItemProps & ListItemLayoutProps, {}> { }
  export class ListItemAction extends React.Component<ListItemActionProps, {}> { }
  export class ListItemActions extends React.Component<ListItemActionsProps, {}> { }
  export class ListItemContent extends React.Component<ListItemContentProps, {}> { }
  export class ListItemLayout extends React.Component<ListItemLayoutProps & ListItemContentProps & ListItemActionsProps, {}> { }
  export class ListItemText extends React.Component<ListItemTextProps, {}> { }
  export class ListSubHeader extends React.Component<ListSubHeaderProps, {}> { }

  export {
    ListCheckboxTheme,
    ListDividerTheme,
    ListItemActionsTheme,
    ListItemActionTheme,
    ListItemContentTheme,
    ListItemLayoutTheme,
    ListItemTextTheme,
    ListItemTheme,
    ListSubHeaderTheme,
    ListTheme,
  };
}
```

This should be enough to satisfy TS.
It was also my first time writing definitions by hand (usually tsc does the job automatically), so sorry once again for causing problems :disappointed:

newash:
@katzoo thanks for the reply, but it turned out to be a not-so-serious issue. For the time being I can use v1.3.3 without any problem.

I just wanted to make notes on the existing issues, so they can be fixed in the future. 😉

PR #1071, opened 2016-12-19
fix few missing propTypes

katzoo:
A few fixes for props missing in propTypes + markdown updates.


components/list/ListItemContent.js
```diff
@@ -34,8 +36,9 @@ const factory = (ListItemText) => {
 
     render () {
       const {children, caption, legend, theme} = this.props;
+      const contentType = this.getType();
```
katzoo: this is not really a proptype change, but it just felt super weird to see the same function call twice on the same line

javivelasco:
Thanks!

PR #1074, opened 2016-12-21
Fix typeof check in isBrowser function

mikkom:
This change fixes a small issue with the isBrowser check. Rather than checking for undefined we should check for the string 'undefined', as the typeof operator always returns a string indicating the type of the operand.


javivelasco:
Thanks!

PR #1082, opened 2016-12-25
restructure ts definitions

panjiesw:
* Create definitions for each raw components
* Make exports of index.d.ts of each components mimic the index.js

This pull, while make the source `components` directories a bit cramped, will make developers who work with typescript be able to import raw components as per doc says with javascript. 

If it's okay with you, please consider merging this as it'll be useful for those working with typescript and wants to extend the raw components, (including me :smile:)

If this get merged, I will do the same for the `css-next` branch.


panjiesw:
I just get my hands on the `css-next` branch and `react-toolbox@2.0.0-beta.4`. And I think this restructure of ts definition is needed because with the `react-toolbox-themr`, we need to import the components without their bundled css, which is not possible right now for typescript developers.

If the `.d.ts` files are too intrusive to the source code, we can place them in a separate directory and copy them to the `lib` dir at build time. Something like this:

```
├── components
├── docs
├── spec
├── tools
└── typings
```

and in typings dir:

```
typings
├── app_bar
│   ├── AppBar.d.ts
│   └── index.d.ts
├── autocomplete
│   ├── Autocomplete.d.ts
│   └── index.d.ts
├── avatar
│   ├── Avatar.d.ts
│   └── index.d.ts
├── button
│   ├── BrowseButton.d.ts
│   ├── Button.d.ts
│   ├── IconButton.d.ts
│   └── index.d.ts
...
```

@javivelasco what do you think?

javivelasco:
@panjiesw Actually I prefer them in the same component directory. This way the relationship is very clear

panjiesw:
Alright then i'll make another pull cause I messed up the tree

Issue #1084, opened 2016-12-25
Tabs pointer causes right arrow to appear when resizing if last tab is selected


raulmatei:
Hi,

I have a uses-case where tabs are divided into two sections: **left tabs** & **right tabs**. I've chosen to go with a single Tabs instance and add a disabled tab between sections as a separator: 
`[Tab1]` `[tab-separator(flex-grow-1)]` `[Tab2]` `[Tab3]` (see below). 

I have to do it this way because I want to have the tabs scrolling functionality in place. 

All's fine but there are some issues:
1. if pointer animation is enabled - it causes the scroller right arrow to appear and the pointer goes crazy, it sometimes updates it's position and sometimes it doesn't
2. if I disable animation - the scroller arrow issue is solved, but there are still issues with updating pointer position sometimes

![tabs-pointer-issue](https://cloud.githubusercontent.com/assets/1718341/21472844/0d6567d8-caf4-11e6-8df4-f520c51ab1b0.gif)

Wouldn't it be better to have a pseudo element (`::after` styled accordingly) or even a `bottom-border` for the active tab item when animation is disabled? This will reduce the number of issues related to updating the position, at least for this mode. Also, the position of the pointer element seem to trigger the scroller issue only when animation is active.

Thanks,
Raul.

raulmatei:
I've had a deeper look into the issue and found out that the pointer element is for sure triggering this issue. That's because when resizing, the pointer element position is not immediately updated so that  this check `nav.scrollWidth > nav.clientWidth && (nav.scrollLeft + nav.clientWidth) < nav.scrollWidth` from `updateArrows` returns `true`, and that's why the right arrow appears when resizing.

I've tried to find a solution, and I think the best way to solve it is to move the pointer element out of `nav`, and wrap it into it's own container, the HTML structure should look like this:

```html
<div class="tabs" data-react-toolbox="tabs">
  <div class="navigationContainer">
    <nav class="navigation">
      <label class="label" data-react-toolbox="tab" id="0"></label>
      <label class="label" data-react-toolbox="tab" id="1"></label>
      <label class="label" data-react-toolbox="tab" id="2"></label>
      <label class="label" data-react-toolbox="tab" id="3">></label>
    </nav>
    <div class="pointerWrapper">
      <span class="pointer" style="..."></span>
    </div>
  </div>
  <!-- tabs content removed to reduce noise  -->
</div>
```

Visually nothing should change:
[...]

PR #1087, opened 2016-12-27
Fix tabs pointer position

raulmatei:
This is the first draft, as described in https://github.com/react-toolbox/react-toolbox/issues/1084#issuecomment-269249082 it changes the HTML structure and the CSS + adds a new method of computing the pointer position. I'd love to find a better solution without another `ref` but I have no other ideas now.

This brings a small issue, because the pointer was moved out of the `nav`, it now appears over the scroller buttons:
image


javivelasco:
I've merged the other approach! Thanks :)

PR #1088, opened 2016-12-27
Fix right arrow appearing when window is re-sized

raulmatei:
As per https://github.com/react-toolbox/react-toolbox/issues/1084#issuecomment-269317228 and the previous pull request https://github.com/react-toolbox/react-toolbox/pull/1087 I've decided to have a closer look on how updateArrows sets the `arrows.right` to `true`. Here's another solution which is less invasive and solves the issue with right arrow appearing on `fixed` tabs when resizing and the last tab is selected.


javivelasco:
Looks like a good solution! Thanks!

Issue #225, opened 2015-12-20
Accesible Components


javivelasco:
This issue will be used to track the progress on adding accessibility attributes to the components. Also, these improvements should take into consideration keyboard navigation through the items. Although it's needed for some components, there are already radio buttons, switch, autocomplete, dropdowns, etc that supports keyboard events.

@nikgraf provided [this link](http://www.w3.org/TR/wai-aria-practices) which I think is a good reference to start from.

nathanmarks:
@javivelasco https://github.com/rackt/react-a11y

javivelasco:
Amazing @nathanmarks, didn't know about it

suancarloj:
Any idea when the dropdown will be accessible with keyboard?

javivelasco:
I was planning to do work on this during next weekend. I already have a `Chip` component to decouple autocompleted fields from the autocomplete itself. Also, I'd make accesible with the keyboard in that feature. If this is going to change there no point on improving it for now!

cometta:
```
<List shit="damn" role='presentation'  selectable ripple >
<ListSubHeader caption='Menu' />
<ListItem role='presentation' tabIndex='0'
caption='Home'
/>
</List>
```
Can help to check why` <list>` and` <ListItem>` filter out `role=`  and `tabIndex` ?

javivelasco:
I don't understand your comment 😕

cometta:
Sorry, let me rephrase. What I meant was, I unable to add accessibility tags to `<List> `and `<ListItem> `component. I can't even add `'id=...'`  . On the other hand, I tested with `<FontIcon> `component and there is no issue.

javivelasco:
Yeah, this is happening because not every component is spreading every passed prop. It would be nice to have PRs when this happens because as in other many cases, this can be discovered with usage

DevAlien:
@javivelasco any news on the dropdowns?

javivelasco:
Not really. I'm migrating to a [different theming approach](https://github.com/react-toolbox/react-toolbox/tree/react-css-themr-migration) and I can't dedicate more time for now.

ab712:
I set aside some time this morning to work on this today but I'm not sure how to drive out the changes from tests. It looks like the specs are creating instances of the component and only fail if an error happens on mounting the component vs doing assertions on behavior effects. Would love to start contributing to this project but I'm not sure how to test new features.

felixzapata:
[...]

PR #1094, opened 2016-12-30
feat(a11y): improve accessibility for iconMenu component

felixzapata:
Hi, this pull request is related with #225. In that issue it says that it should be done from #666 but, because it has not been merged yet, I think that this pull request can be a good idea to take a look, at least, to the changes required to improve the accessibility of the `iconMenu` component. 

The accessibility improvements are:
+ add aria attributes to establish relationship between the icon and the menu, and also the visibility of the menu.
+ set the focus when the menu is opened. Important note: it is possible that the `MenuItem` component should be fixed too, because the elements inside it, are not `focusables`.. they are just `span` elements.
+ add support when the user press the escape key.

Regards.

### DIFF
--- a/components/button/IconButton.js
+++ b/components/button/IconButton.js
@@ -9,6 +9,7 @@ const factory = (ripple, FontIcon) => {
   class IconButton extends Component {
     static propTypes = {
       accent: PropTypes.bool,
+      ariaControls: PropTypes.string,
       children: PropTypes.node,
       className: PropTypes.string,
       disabled: PropTypes.bool,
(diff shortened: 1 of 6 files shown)


javivelasco:
I'm closing this one because it has the same issues as #1105. This component is adding a lot of event listeners and it should be updated. Also, you might be interested about knowing that this component is going to change a lot very soon because it's one of the conflictive ones so maybe it's worthy to review the implementation and send a different PR when you have time. Thank you for this hard work :)

PR #1098, opened 2017-01-01
Added proper type definitions for dialog actions so typescript users …

owais:
…can pass on props to the generated buttons


owais:
@javivelasco bump :)

javivelasco:
Thanks!! :)

PR #1099, opened 2017-01-02
Preview of what typescript would feel like

owais:
Quick and dirty preview. Probably `Handler<ReactNode>` is not the best thing to do but the the idea is to showcase what writing in Typescript would feel like.


javivelasco:
This looks pretty good @owais but honestly I don't have time enough to get familiar with TS, migrating components and keep up with all other stuff listed in the Roadmap. :( What we do need is someone officially responsible from maintaining the typings files. Thanks for the approach, again, it looks great but it's not the moment on my side :(

PR #1101, opened 2017-01-02
feat(a11y): alternative text for the FontIcon

felixzapata:
- add attribute to set an alternative text to the icon.
- hide the text inside the component used by Material to render the icon.
- remove unused attribute `children`. This attribute had the same description as the `value` attribute in the README file.

It is related with #225.


javivelasco:
Thank you!!

PR #1102, opened 2017-01-02
feat(a11y): alternative text for Avatar component

felixzapata:
Allow to add an alternative text to the icon or image. It is related with #1100 and #1101 and #225.


javivelasco:
Cool! Thanks!

PR #1105, opened 2017-01-03
feat(a11y): improve accessibility for the SnackBar component

felixzapata:
+ set the focus when the message is opened.
+ return the focus to the element which has opened the message.
+ show focus style when you focus over the button inside the message.

Related with #225.

### DIFF
--- a/components/snackbar/Snackbar.js
+++ b/components/snackbar/Snackbar.js
@@ -19,6 +19,7 @@ const factory = (Overlay, Button) => {
       ]),
       onClick: PropTypes.func,
       onTimeout: PropTypes.func,
+      opener: PropTypes.object,
       theme: PropTypes.shape({
         accept: PropTypes.string,
         active: PropTypes.string,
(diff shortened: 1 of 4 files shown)


components/snackbar/Snackbar.js
```diff
@@ -39,15 +40,31 @@ const factory = (Overlay, Button) => {
     }
 
     componentWillReceiveProps (nextProps) {
+      const listenerTransition = () => {
+        this.setFocus();
+        this.refs.snackbar.removeEventListener('transitionend', listenerTransition);
+      };
+      if (nextProps.active) {
+        this.refs.snackbar.addEventListener('transitionend', listenerTransition);
```
javivelasco: Isn't this adding tons event listeners? I think it should be checked if the listener is already added. May you look into it?

felixzapata:
ok, I will review it

components/snackbar/Snackbar.js
felixzapata: ok, I will review it

SleeplessByte: Instead, wouldn't it be better to add to `componentDidMount` and `componentWillUnmount`, register and unregister once, and simply handle it in the callback using props?.

PR #1110, opened 2017-01-04
Allow to change 'source' for single 'Autocomplete'

HEm3R:
It is now able to dynamically change source for single value
Autocomplete, for example in 'onInput' handler.

In case you change the `source` property for Autocomplete in `onQueryChange` handler:

```
queryChanged(value) {
  this.setState({employees: this.loadEmployees(value)});
}
```

for single value `Autocomplete`:

```
<Autocomplete
  multiple={false} label="Employee" name="employee" source={this.state.employees}
  value={this.state.employee} onInput={this.queryChanged.bind(this)}
/>
```

The value in `Input` component keeps empty or - in case you select value - cannot be changed.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -74,14 +74,6 @@ const factory = (Chip, Input) => {
      isValueAnObject: false
    };
 
-   componentWillReceiveProps (nextProps) {
-     if (!this.props.multiple) {
-       this.setState({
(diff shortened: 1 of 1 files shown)


HEm3R:
I have started with React Toolbox couple of days ago, so if this PR violates any rules or there is reason why it behaves as now, please let me know. Thanks!

radeklos:
👍

PR #1115, opened 2017-01-06
Remove diacritic signs in query and suggestions

radeklos:
You don't get any suggestion when they contains diacritic sings. This PR normalises query value and suggestions.


javivelasco:
Thanks! Note this will be rewritten soon :)

PR #1120, opened 2017-01-07
Enable dynamic source for single value Autocomplete

HEm3R:
It is now able to dynamically change source for single value
Autocomplete, for example in 'onInput' handler.

In case you change the `source` property for Autocomplete in `onQueryChange` handler:

```
changeSource(value) {
  this.setState({employees: this.loadEmployees(value)});
}
```

for single value `Autocomplete`:

```
<Autocomplete
  multiple={false} label="Employee" name="employee" source={this.state.employees}
  value={this.state.employee} onInput={this.changeSource.bind(this)}
/>
```

The value in `Input` component keeps empty or - in case you select value - cannot be changed.

This issue with simple autocomplete is fixed by this PR, example is added to docs.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -34,6 +34,7 @@ const factory = (Chip, Input) => {
      onChange: PropTypes.func,
      onFocus: PropTypes.func,
      onQueryChange: PropTypes.func,
+     query: PropTypes.string,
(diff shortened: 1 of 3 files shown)


HEm3R:
Original buggy PR was here: https://github.com/react-toolbox/react-toolbox/pull/1110

PR #1121, opened 2017-01-07
Fix error if list get null object

mxck:
Simple example:
render() { const a = null; return <List>{{a}}</List>; }

Get 'Uncaught TypeError: Cannot read property 'type' of null'


javivelasco:
This is already fixed! Thanks for the patch

PR #1122, opened 2017-01-07
Add suggestionMatch option to disable filtering

HEm3R:


javivelasco:
Looks reasonable, thanks!

Issue #1140, opened 2017-01-14
v1 typescript typings bug - index.js export Ripple


beckend:
Module '"node_modules/react-toolbox/lib/ripple/index"'
 has no exported member 'Ripple'.

And it truth, it does not.
Reading the src this is the only export of [default](https://github.com/react-toolbox/react-toolbox/blob/master/components/ripple/index.js#L4);

panjiesw:
How do you import it?

```javascript
// It's supposed to be used like this
import Ripple from 'react-toolbox/lib/ripple'
// not this one
import {Ripple} from 'react-toolbox/lib/ripple'
```

But yeah, it wouldn't hurt to also provide the named export `Ripple`

beckend:
Related to https://github.com/react-toolbox/react-toolbox/issues/1125 if he meant v1.

PR #1141, opened 2017-01-14
v1, Typescript typings fix Ripple index export.

beckend:
[issue 1140](https://github.com/react-toolbox/react-toolbox/issues/1140).

### DIFF
--- a/components/ripple/index.js
+++ b/components/ripple/index.js
@@ -1,4 +1,5 @@
 import rippleFactory from './Ripple.js';
 import theme from './theme.scss';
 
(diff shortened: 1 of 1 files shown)


Pajn:
You should probably fix the typings file [here](https://github.com/beckend/react-toolbox/blob/a37b082cd32156fd2d94c4eae8535ba0b8198f8e/index.d.ts#L25) instead of changing the API of the project

beckend:
Well either is fine, but the API does not get changed by this change.

PR #1153, opened 2017-01-18
Fixed #894 by adding a disableScrolling prop

### LINKED ISSUE #894: dialog size
I have a dropdown in a dialog box.  When the drop down is activated, it's cut off because the dialog box isn't big enough.

I tried putting a div with height:100% as a container inside my dialog box.  Didn't work.
I tried styling the dialog box directly with height:100%.  Didn't work.

```
            <Dialog
                actions={this.actions}
                active={this.props.active}
                onEscKeyDown={this.props.onToggle}
                onOverlayClick={this.props.onToggle}
                title='Edit Task' 
                type='fullscreen'
                <p>placeholder</p>
                <Dropdown
                    auto
                    source = {this.dropDownSource}
                    value = {this.state.dropDownValue}
                    onChange = {this.handleDropDownChange}
                />
            </Dialog>

```

2sComplement:
This seems like a good compromise between having scrolling enabled by default, and allowing the ability to disable if required.

### DIFF
--- a/components/dialog/Dialog.js
+++ b/components/dialog/Dialog.js
@@ -17,6 +17,8 @@ const factory = (Overlay, Button) => {
       [props.theme.active]: props.active
     }, props.className);
 
+    const bodyClassName = props.disableScrolling ? props.theme.body : classnames(props.theme.body, props.theme.scroll);
+
     return (
       <Overlay
         active={props.active}
@@ -27,7 +29,7 @@ const factory = (Overlay, Button) => {
         onMouseUp={props.onOverlayMouseUp}
       >
         <div data-react-toolbox='dialog' className={className}>
-          <section role='body' className={props.theme.body}>
+          <section role='body' className={bodyClassName}>
             {props.title ? <h6 className={props.theme.title}>{props.title}</h6> : null}
             {props.children}
           </section>
@@ -47,6 +49,7 @@ const factory = (Overlay, Button) => {
     active: PropTypes.bool,
     children: PropTypes.node,
     className: PropTypes.string,
+    disableScrolling: PropTypes.bool,
     onEscKeyDown: PropTypes.func,
     onOverlayClick: PropTypes.func,
     onOverlayMouseDown: PropTypes.func,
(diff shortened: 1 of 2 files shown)


javivelasco:
Agree! Thanks :)

2sComplement:
@javivelasco I know npm is currently on v2.0.0-beta, but is there a way to get this fix in an npm version 1.3.5?

javivelasco:
Limited time on my side :( If you really need it, you can open a PR with the fix to master and I'll be happy to release a patch

2sComplement:
I was hoping to update [react-toolbox bindings](https://github.com/2sComplement/fable-react-toolbox) for [Fable](http://fable.io/), so a patch would help. As a side-note, you may be interested in that project anyway.

Going back to my original question, I'm confused. This PR was to master, no?

javivelasco:
Sorry, I'm stupid! I thought it was a fix merged for the current `dev`. Of course I can release a new patch from master. Sorry about the confusion! Gonna check the project, looks interesting! thank you!

2sComplement:
No worries, I can PR into ``dev`` if you want. I figured at some point you'll be merging into master anyways 😄.

landabaso:
Sorry to pop this up again.

I believe it should have been:

```
.scroll {
   overflow-y: scroll;
 }
```

instead of current:
```
.scroll {
   overflow: hidden;
 }
```
Is it there any reasoning behind I'm missing?
As it is now, it is not possible to scroll.

2sComplement:
@landabaso you are correct. I had originally coded the fix on my Windows machine but the build seems to fail on Windows so I switched to Linux to test and commit the fix and 💥. I just submitted #1199 which should fix the problem.

landabaso:
Thanks! There's yet another issue that we should consider.
I'm not sure using vh for implementing Dialogs is a good idea.

Note that when the Dialogs are large enough to overflow, then the bottom part of Dialogs will be hidden behind the toolbars in iOS and (now also) Chrome in Android. The reason is the way vh is implemented in those devices.

You can read more about that issue here:

https://nicolas-hoizey.com/2015/02/viewport-height-is-taller-than-the-visible-part-of-the-document-in-some-mobile-browsers.html

Since iOS and Android amount for a GREAT percentage of the browsers out there I believe it would be appropriate to consider alternatives.

javivelasco:
Fix merged. I'm releasing `master` in a couple of days... so busy!

PR #1154, opened 2017-01-18
Typescript: document  Input methods

Strate:
As described in the docs:

blur used to blur the input element.
focus used to focus the input element.

http://react-toolbox.com/#/components/input


javivelasco:
Cool! Thanks!

Issue #1118, opened 2017-01-06
Create a test / spec for typescript definition


panjiesw:
As the lib is not originally authored in typescript, the definition is written manually by hand, thus may introduce an incomplete definition, or typo, error, etc for the ts users.

Should we make some kind of type checking for the definition into the build/test step? We can do it like the type registration in [DefinetlyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), where there is a spec only typescript file importing all the components, which purpose is only for type checking. 

For example we can define the `spec` directory as a typescript project (put a `tsconfig.json` into this), then we write a ts file importing all the component, then run the `tsc` compiler against it. This way any contribution to typescript definition of `react-toolbox` will be either type checked by IDE/editor or the `tsc` compiler.

javivelasco:
This would be perfect. Would you create a PR for this? It would be very helpful specially for me since I bever used Typescript

panjiesw:
Sure, I'll get into this soon. Should I include this in #1114 ?

javivelasco:
I'd make itnin a separate PR, I'm merging that one now

Issue #935, opened 2016-11-06
Add "export" to Props interface definition in all component index.d.ts files


uds:
Having standard Props interface defined as private makes it impossible to reuse the Props definition when creating your own custom version of the component.

javivelasco:
Anybody helping with this?

nidu:
Is anybody doing it? I'd like to slightly extend some components in my app and reusing existing props would be very convenient.

This sounds quite simple. Just exporting props won't break anything and actually we're not exposing more than already exposed because Props are a part of API

javivelasco:
Not that I know @nidu!

PR #1163, opened 2017-01-20
Typescript definitions validation

panjiesw:
This is a follow up attempt of the one described in #1118 .

- [x] Add typescript compiler and types dependencies.
- [x] Add task to package.json to validate definitions (without emitting the compiled result).
- [x] Add components spec written in typescript in `spec/ts` directory.

The third one is a going to be a while. I think it's necessary to write them so anyone contributing any ts definitions can try to use it there. But I'm open to any other suggestion of how to achieve the intention.


javivelasco:
How is this going? What do you think about #935?

panjiesw:
@javivelasco slow, unfortunately... :frowning_face: 

#935 is already implemented in `1.3.4` actually. The props and theme interfaces are all already exported in index.d.ts when I did the restructure.

javivelasco:
Don't worry, take your time! I'm closing that one then

Kikobeats:
IMHO would be better move this specific typescript rules in a separate and complementary package, like `react-toolbox-definitions`

panjiesw:
@javivelasco do you still interested in this? Or should we separate the typescript stuff outside this repo as @Kikobeats said?

panjiesw:
@javivelasco it's done now, it can be merged if you still have interest in this

javivelasco:
Gotta check it out in more detail, patience! :)

stereokai:
@panjiesw  @javivelasco when this is merged, would it be possible to use react-toolbox with typescript?

panjiesw:
You can use `react-toolbox` with typescript just fine right now both in stable `1.3.4` and beta `2.0.0-beta.6`.

This PR is more of an enhancement for the development of `react-toolbox` and its typescript definitions, so that typescript users will have more pleasant experience when creating app with those wonderful tools (react-toolbox and typescript) :smile:

stereokai:
I see. Thanks a lot :)

seivan:
I second to add this directly to this repo so it just works out of the box instead of having to add dev-dependencies (@types/react-toolbox) 
I noticed there were test cases to test existing type definitions. Is there a way to setup a linter on the CI to ensure that future components also have typescript support?

Issue #1155, opened 2017-01-19
Tabs are not rendered [2.0.0-beta4]


olegstepura:
Fully copying the full Tabs example component from here: http://react-toolbox.com/#/components/tabs with latest beta release

I see this (no tabs at all):
![tabs](https://cloud.githubusercontent.com/assets/534510/22099644/dd83ff9e-de34-11e6-9b6c-52ad8d8a1b67.png)

Didn't you change the props? Or is it a bug?

javivelasco:
Maybe there is a bug in that specific version. Probably there were missing exports. I've tried the latest released beta and it works for me. Please try it out and reopen if it doesn't work! Thanks for reporting :)

olegstepura:
Hi! It's the same in `react-toolbox@2.0.0-beta.4`. Please reopen.

javivelasco:
Can anybody else confirm this issue? I've tried and it works for me :/

olegstepura:
I'll try to create a test repo

javivelasco:
Ok! Thanks :)

olegstepura:
Here is my test case for Tabs (`tabs-test` branch):
https://github.com/olegstepura/react-toolbox-example/tree/tabs-test

For me the resulting picture is still the same (tested in Firefox and Chrome on Windows)

javivelasco:
That wasn't easy. The issue is that React Hot Loader proxies components so the `type` of the component is no longer recognized. This happens because `Tabs` needs to recognize its children to figure out how should they be rendered. We [added recently](https://github.com/react-toolbox/react-toolbox/pull/1164) the ability to override `isComponentOfType` to keep using Hot Loader with react-toolbox. There are instructions in the same PR. What this fix does is to use `isComponentOfType` implementation to allow overriding it.

olegstepura:
So I just need to wait for #1165, correct?

javivelasco:
I've just released 😵 It's going to work for the `Layout` but not for `Tabs` because the fix is not included in the bundle. It will be for the next release!

SleeplessByte:
@javivelasco please update Tabs.d.ts in `react-toolbox@2.0.0-beta.5`

It refers to `TabTheme` but doesn't import it (from `Tab.d.ts`)

javivelasco:
Done @SleeplessByte, coming with the next beta

olegstepura:
When do you think the next beta will be released?

javivelasco:
Released!

[...]

Issue #1152, opened 2017-01-18
[2.0.0-beta1] NavDrawer permanentAt property seem to be ignored


olegstepura:
Hi! When using the 2.0 beta settings `permanentAt="lg"` seem to be ignored even on large width (more than 2000px).

I'm using it the same way as in docs:

```js
import React from 'react'
import { Layout, AppBar, Navigation, Link, NavDrawer, Panel } from 'react-toolbox'
import theme from './Wrapper.css'

class Wrapper extends React.Component {
  state = {
    drawerActive: false,
    drawerPinned: false,
    sidebarPinned: false
  }

  toggleDrawerActive = () => {
    this.setState({ drawerActive: !this.state.drawerActive })
  }

  render() {
    return (
      <Layout>
        <NavDrawer active={this.state.drawerActive}
                   pinned={this.state.drawerPinned} permanentAt="lg"
                   onOverlayClick={this.toggleDrawerActive}>
          <p>test</p>
        </NavDrawer>
        <Panel>
          <AppBar title="Test" leftIcon="menu" onLeftIconClick={this.toggleDrawerActive} theme={theme}>
          </AppBar>
          {this.props.children}
        </Panel>
      </Layout>
    )
  }
}

export default Wrapper
```

mikelgarciaurbina:
Hi, I checked it out, and it's working for me in the 2.0.0-beta.4 😉

javivelasco:
Cool! Thanks @mikelgarciaurbina !

olegstepura:
Hi! I checked the code and it seems like the `permanentAt` property is ignored
https://github.com/react-toolbox/react-toolbox/blob/dev/components/layout/NavDrawer.js#L12
It even has `eslint-disable-line`, I think it's to skip variable usage.

@mikelgarciaurbina how can it work for you? Are you sure you are using beta version?

olegstepura:
@javivelasco can you please take a look at the code of `NavDrawer` if `permanentAt` should work?

javivelasco:
It turns out that the `permanentAt` property is read at the `Layout` component to determine if `NavDrawer` / `SideBar` are pinned. If so, it is controlled with CSS from the `Layout` component. The property has to be ignored in the `NavDrawer` and `SideNav` to allow spreading additional props to the top level element.

olegstepura:
Ok, I will create a test stand ;)

javivelasco:
You can check how is it working in the example from spec folder root component too!

[...]

PR #1164, opened 2017-01-21
Ability to override default isComponentOfType implementation

### LINKED ISSUE #1148: Dependency injection doesn't work with React-hot-loader
Hey, first of all thanks for a great library! Love it.
I'm using `2.0.0-beta.4` version and there's a utility function that helps filter out children of certain types:
```js
function isComponentOfType(classType, reactElement) {
  return reactElement && reactElement.type === classType;
}
```
Approach described here https://github.com/react-toolbox/react-toolbox/issues/957 with `CustomRow` injection doesn't work for me in development environment, because all components are proxied by `React-hot-loader` and strict equality check fails. Is there another way to do it or is it possible to expose an API to change the behavior of `isComponentOfType`? `React-hot-loader` is really popular, it would be great to have `react-toolbox` compatible with it.

valerybugakov:
Fixes #1148 
Let me know if any fixes required.

### DIFF
--- a/components/index.js
+++ b/components/index.js
@@ -1,4 +1,5 @@
 import './utils/polyfills'; // Import polyfills for IE11
+export { overrideComponentTypeChecker } from './utils/is-component-of-type';
 export AppBar from './app_bar';
 export Autocomplete from './autocomplete';
 export Avatar from './avatar';
--- a/components/utils/is-component-of-type.js
+++ b/components/utils/is-component-of-type.js
@@ -1,9 +1,33 @@
+let customChecker;
+
+/**
+ *  Sets customChecker which will be used for all components.
+ *
+ * @param providedChecker {Function} - Checker function
+ */
+export function overrideComponentTypeChecker (providedChecker) {
+  customChecker = providedChecker;
+}
+
 /**
  * Returns true if the provided element is a component of the provided type.
  *
  * @param classType {ReactElement class} - the class of a React Element
  * @param reactElement {ReactElement} - any React Element (not a real DOM node)
  */
-export default function isComponentOfType (classType, reactElement) {
+export function defaultChecker (classType, reactElement) {
   return reactElement && reactElement.type === classType;
+}
+
+
+/**
+ * Executes customChecker if it's set or defaultChecker.
+ *
+ * @param classType {ReactElement class} - the class of a React Element
+ * @param reactElement {ReactElement} - any React Element (not a real DOM node)
+ */
+export default function isComponentOfType (classType, reactElement) {
+  return customChecker
+    ? customChecker(classType, reactElement)
+    : defaultChecker(classType, reactElement);
 }


javivelasco:
How do you use it? I see there is a variable defined on the global scope of the module but are you overriding through window? The usage should be defined as well. I don't like too much the function to depend on an external variable definition, maybe there is a way to achieve a similar behavior using something more explicit like `env` variables. This way you transform the source specifying wether you want to use the default checker when you run the transpiler.

valerybugakov:
Hey, the variable is defined not in global scope but in scope of the module `is-component-of-type.js`. The usage is pretty simple. In your root React file just import `overrideComponentTypeChecker` and pass there your checker function.  I don't see a way to change the default behaviour without changing an external value. Here's an example from my app with `React-hot-loader`: 
```js
import React from 'react'
import { render } from 'react-dom'
import { AppContainer } from 'react-hot-loader'
import { overrideComponentTypeChecker } from 'react-toolbox'
import store from 'redux/configureStore'
import Root from 'components/Root'

overrideComponentTypeChecker((classType, reactElement) => (
  reactElement && (
    reactElement.type === classType ||
    reactElement.type.name === classType.displayName
  )
))

const throwError = ({ error }) => { throw error }
render(
  <AppContainer errorReporter={throwError}>
    <Root store={store} />
  </AppContainer>,
  document.getElementById('root'),
)

if (module.hot) {
  module.hot.accept('components/Root', () => {
    System.import('components/Root').then(RootModule => {
      const UpdatedRoot = RootModule.default

      render(
        <AppContainer errorReporter={throwError}>
          <UpdatedRoot store={store} />
        </AppContainer>,
        document.getElementById('root'),
      )
    })
  })
}
```

Let me know what do you think. I didn't get an idea about `env` approach. How will you provide the checker function then?

javivelasco:
No your right, I think it makes sense. Now that the example is documented in the same PR we can merge it. Looks good, thanks! :)

olegstepura:
Such documentation should be in README.
Currently it can be just a link to this PR discussion to allow all users with

olegstepura:
I suggest this change:

## Usage with *react-hot-loader v3*

Since [react-hot-loader v3](https://github.com/gaearon/react-hot-loader/pull/240) babel patch leads to issues with some components (see #1152, #1155) to allow hot reloading in dev mode you need to define your own type checker in your main entry point (see #1164 for details):

```js
import "react-hot-loader/patch"
import React from 'react'
import { overrideComponentTypeChecker } from 'react-toolbox'

overrideComponentTypeChecker((classType, reactElement) => (
  reactElement && (
    reactElement.type === classType ||
    reactElement.type.name === classType.displayName
  )
))
```

javivelasco:
Looks good! PR?

olegstepura:
I was thinking about this addition. And here is my thoughts.

1. I don't like the idea to force usage of `overrideComponentTypeChecker` in main entry point. 
    - Maybe we can just add this code (` ||  reactElement.type.name === classType.displayName`) to the default type checker. Not sure if this can lead to issues.
    - Or maybe we can use a switch to detect react hot loader and use custom code. E.g.
    ```js
    export function defaultChecker (classType, reactElement) {
      if (typeof global.__REACT_HOT_LOADER__ !== 'undefined') { // or if (process.env.NODE_ENV !== 'production')
        return reactElement && (
          reactElement.type === classType ||
          reactElement.type.name === classType.displayName
        )
      } else {
        return reactElement && reactElement.type === classType
      }
    }
    
        
    ```
2. If we cannot avoid focing users to add this extra code in main file (why?), maybe we can provide a ready method for not to write this copy-paste. Like `overrideComponentTypeChecker(hotLoaderTypeChecker)` or just `setComponentTypeCheckerForHotLoader()`

javivelasco:
It shouldn't create any issues. Do you want to PR this fix? Makes total sense

valerybugakov:
It would be great to still have an ability to change `defaultTypeChecker` by hand or to switch it into `displayName` mode manually. I think there may be other cases except of `React-hot-loader` compat.

olegstepura:
@javivelasco which version do you want as a PR?

olegstepura:
Ok, my vision is as such:

```js
import global from 'global'

let customChecker

export function overrideComponentTypeChecker (providedChecker) {
  customChecker = providedChecker
}

export function defaultChecker (classType, reactElement) {
  return reactElement && reactElement.type === classType
}

export function defaultCheckerWithDisplayNameCheck(classType, reactElement) {
    return reactElement && (
      reactElement.type === classType ||
      reactElement.type.name === classType.displayName
    )
}

export default function isComponentOfType (classType, reactElement) {
  if (customChecker) {
    return customChecker(classType, reactElement)
  } else if (typeof global.__REACT_HOT_LOADER__ !== 'undefined') {
   return defaultCheckerWithDisplayNameCheck(classType, reactElement)
  } else {
    return defaultChecker(classType, reactElement)
  }
}
```
please comment if you accept this (with an extra dep 'global') and if you accept code without semicolons  . I will provide a PR afterwards.

javivelasco:
To me it looks ok and functional. But it needs to be adapted to the code style defined by eslint which basically it's airbnbs (*with* semicolons)

olegstepura:
ok, I'll do that

javivelasco:
Actually I'm not sure why shouldn't we just make the check in the same function.

olegstepura:
please decide ;)
Not sure, but issues could raise out of minification (if comparing `type.name` would lead to false positives)
Related: [hot-loader issue](https://github.com/gaearon/react-hot-loader/issues/304) and [react-devtools issue](https://github.com/facebook/react-devtools/issues/239)

javivelasco:
I'm not sure about this anyway. I was trying the solution in a project with react-hot-loader and it didn't work nicely. Maybe we can just keep the ability to override the function so anybody can write its own.

olegstepura:
Maybe point 2 from https://github.com/react-toolbox/react-toolbox/pull/1164#issuecomment-275062357 is what we need? E.g. `overrideComponentTypeChecker(hotLoaderTypeChecker)`

PR #1187, opened 2017-01-28
Update Input with Icon to follow Material Spec.

izakfilmalter:
image

As you can see, the spec has a 20dp spacing between the icon, and the input. I added `$input-icon-right-space` to allow this to be customized and defaulted it to the 20dp in the spec.

image

Link to the spec:
https://material.io/guidelines/components/text-fields.html#text-fields-single-line-text-field


izakfilmalter:
Let me know if you want a different variable name.

javivelasco:
Cool, thanks!

Issue #1158, opened 2017-01-20
Component naming for 2.0


olegstepura:
I suggest changing names of components inside `component` folder to camel case.
Same is done in [material-ui](https://github.com/callemall/material-ui/tree/master/src). It look better when importing separate modules (and with webpack if I understand it correctly, you should better use separate component imports instead of importing required components from main index.js to compile only required for production)
It would look much better to instead of this:
```js
import AppBar from 'react-toolbox/lib/app_bar';
```
to use this:
```js
import AppBar from 'react-toolbox/lib/AppBar';
```

Not sure if with webpack 2 with tree-shaking you will still need to import only required modules.

javivelasco:
I *TOTALLY* agree on this.

Reanmachine:
@javivelasco One thing that I've noticed using `react-toolbox` 2.0 with CRA is that we have a much more verbose inclusion pattern:

Before:

```
import {
  Card,
  CardTitle,
  CartActions
} from "react-toolbox/lib/card"; // or from "react-toolbox"
```

After:
```
import Card from "react-toolbox/lib/card/Card";
import CardTitle from "react-toolbox/lib/card/CardTitle";
import CardActions from "react-toolbox/lib/card/CardActions";
```

One thing I was thinking of as a possibility for improving this procedure was a route that `redux-forms` went with `immutable` bindings.

Perhaps this could be a reality:

```
import {
    Card,
    CardTitle,
    CartActions
} from "react-toolbox/unstyled";
```

Or perhaps this for setups that don't have access to treeshaking:
```
import {
   Card,
   CardTitle,
   CardActions
} from "react-toolbox/lib/Card/unstyled";
```

What are your thoughts? I think with the 2.0 beta if we're re-thinking the import strategy for the styled-components approach it's worth looking at the usability & tree shaking structure for the unstyled + `react-toolbox-themr` approach.

javivelasco:
@Reanmachine I agree. It's very messy to mix up imports with style and without style. My primary intention is to move the implementation to **agnostic components** as shown in [this tweet](https://twitter.com/javivelasco/status/825048471480696833) with some WIP.

This migration will allow to move components to different folders so there would be some different packages like:

- Styled Components
- Styletron
- CSS Modules
- Unstyled

That's a good moment to change names trying to keep backwards compat in the same structure as it is now. I'd like to keep #1189 away from merging until the first migration is finished and we can get feedback on the best organization for the package.

[...]

PR #1189, opened 2017-01-29
Rename components folders

rubenmoya:
Related to #1158


javivelasco:
There is going to be a massive refactor with the POC of agnostic components so I'm better closing. Thanks for the PR anyway! :)

PR #1199, opened 2017-01-31
Use correct overflow rule in scroll style

2sComplement:
Fixing bug in #1153, which contained a fix for #894.


javivelasco:
Thanks!

PR #1208, opened 2017-02-01
Fixes #1086

### LINKED ISSUE #1086: Dropdown stopped working correctly when opened up
Here is how it currently behaves for me:
dropdown-bug

The reason is this commit: https://github.com/react-toolbox/react-toolbox/commit/2f5e0f5f358db8119913204814596711fd62cefd

`onFocus` is called when I click on the dropdown which calls `open`.

Rolling back to 1.2.5 helped

jmar777:
This PR fixes the issue described in #1086.

The bug results from `handleFocus()` attempting to call `open()` on a Dropdown instance that is already open. In most cases, this is a no-op, but some window size + page layout combinations cause a new value for `this.state.up` to be calculated, which results in the Dropdown items shifting on screen, and the subsequent click event not being dispatched to the intended Dropdown item.

This could alternatively be addressed by checking the `active` state inside of `handleFocus()`, but it seems to me that performing the check in `open()` will catch any similar corner cases where the Dropdown items would otherwise shift while already in an open state.

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -81,6 +81,7 @@ const factory = (Input) => {
     });
 
     open = (event) => {
+      if (this.state.active) return;
       const client = event.target.getBoundingClientRect();
       const screenHeight = window.innerHeight || document.documentElement.offsetHeight;
       const up = this.props.auto ? client.top > ((screenHeight / 2) + client.height) : false;


javivelasco:
Thank you!

jmar777:
@javivelasco No problem. Quite literally the least I can after leveraging your awesome project!

Two quick questions: would you also like a PR against the dev branch? Judging from the code, it looks like the same issue would effect it as well. And do you have an ETA for a new 1.x publish?

javivelasco:
Thanks! You can PR if you want, but that component is going to be completely rewritten soon. It will work for beta releases meanwhile, feel free to do so! There is no ETA, sorry :( only objectives are set but the main purpose is to make beta releases perfectly usable and bug free. Its going to be beta still because huge API changes would come...

jmar777:
Awesome, thanks. I'll just point our package.json towards the commit we need, then. 👍

PR #1222, opened 2017-02-07
Move eslint-plugin-babel to devDependencies

gustavohenke:
Had I seen this I would have packed it together with #1214 😊


javivelasco:
Whoops! Thanks!

Issue #1228, opened 2017-02-08
Should Ripple Fx start onMouseUp/onTouchEnd?


landabaso:
I'm not sure about this one.
I've been checking Material Design specifications but could not find anything.

Say you have a view with Buttons. When scrolling it you may tap a button that will trigger the Ripple Fx but the action won't be triggered since it's not an onMouseUp event (you are scrolling).

Similarly on the Desktop, if you click a Button but then move the mouse out of the Button, the Ripple Fx will be triggered but the action won't.

Is a Ripple Fx a hint to the user to indicate that the action will start? If so, then maybe Ripples should be triggered onMouseUp. Opinions?

landabaso:
I just submitted a PR. It does not fix the issue 100% but it helps a bit by stopping Ripple Fx when the user is scrolling on a touch device.

javivelasco:
I'm also struggling with this issue. I think the best could be (could because I didn't tried it out) to trigger the ripple if after the touch start there only if there is no touch move. As you commented, the ripple should not be triggered while scrolling but I have to say that even Polymer's Button behaves this way. I'm currently refactoring the ripple so I will try out some different implementations to try to get it working better

PR #1229, opened 2017-02-08
Deactivate the Ripple Fx on touchmove

landabaso:
It deactivates the Ripple Fx on touch devices when the user is scrolling.
See https://github.com/react-toolbox/react-toolbox/issues/1228 for further discussion.

### DIFF
--- a/components/ripple/Ripple.js
+++ b/components/ripple/Ripple.js
@@ -196,9 +196,13 @@ const rippleFactory = (options = {}) => {
        * @return {Function} Callback function that deactivates the ripple and removes the listener
        */
       addRippleDeactivateEventListener(isTouch, rippleKey) {
-        const eventType = isTouch ? 'touchend' : 'mouseup';
+        const eventType = isTouch ? ['touchend', 'touchmove'] : 'mouseup';
         const endRipple = this.createRippleDeactivateCallback(eventType, rippleKey);
-        document.addEventListener(eventType, endRipple);
+        if (Array.isArray(eventType)) {
+          eventType.map(eventType => document.addEventListener(eventType, endRipple));
+        } else {
+          document.addEventListener(eventType, endRipple);
+        } 
         return endRipple;
       }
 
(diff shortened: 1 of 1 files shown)


javivelasco:
Thanks for the PR! I have a question on this. If the issue is that the ripple is started on touch start, this PR is not fixing it because you would still see the effect, right? I mean, you are deactivating the ripple not only on touchend but also on touch move. That doesn't prevent the ripple from triggering but deactivates the ripple immediately (which is ok too).

Shouldn't we try to prevent the ripple from triggering?

landabaso:
I've been experimenting and did not reach a conclusion yet  @javivelasco.

There are also subtle problems when triggering the Ripple FX on "click" too:

Imagine the case when you click a button that leads to a large DOM reflow/repaint. I've observed this situation is pretty bad for Ripple Fx animations since everything is triggered by React at the same instant (because of the batched updates process inherent to React).

It is also a problem if the "click" closes the Button that triggered the event. Then you would not see Fx.

There's another possible way to trigger the Ripple Fx though. For example, in iOS native apps there's a state change of the style of the Rows in a List. A grey mask is shown so that the user gets the feedback that the Row has been clicked. I've observed that the grey mask appears a few milliseconds after a touchstart if there's not a touchmove. When a user is scrolling then the grey mask does not appear unless you do very slow scrolling movements.

Similarly as in iOS, a possibility could be delaying the Ripple Fx after a touchstart for 100ms (for example) using a setTimeout. It a touchmove/touchcancel happens before the 100ms, then cancel the timeout handler. If a click happens do also cancel the timeout and trigger the Ripple Fx immediately.

These are only ideas for discussion.

javivelasco:
That's what I was thinking about. It shouldn't be a problem if we wait for the next animation frame before painting. In any case to respect the MD guidelines it should happen on mouse down or touch start but indeed we should delay some milliseconds to detect a scroll. I'm including this in a POC I'm pushing this week

landabaso:
Great! Thanks!

components/ripple/Ripple.js
```diff
@@ -196,9 +196,13 @@ const rippleFactory = (options = {}) => {
        * @return {Function} Callback function that deactivates the ripple and removes the listener
        */
       addRippleDeactivateEventListener(isTouch, rippleKey) {
-        const eventType = isTouch ? 'touchend' : 'mouseup';
+        const eventType = isTouch ? ['touchend', 'touchmove'] : 'mouseup';
```
javivelasco: The code would look easier if you return always an array and iterate over it.

```jsx
const eventTypes = isTouch ? ['touchend', 'touchmove']: ['mouseup'];
event.types.forEach(eventType => {
    document.addEventListener(eventType, endRipple);
});
```

dodgecm:
@javivelasco is there any way that I can push this PR along?  should i fork the patch, add the change, and submit a new PR?

javivelasco:
@dodgecm Sure thing, I'm fixing it and including it in the next release

PR #1239, opened 2017-02-13
Invalid appBarIconVisible calculation

### LINKED ISSUE #1231: Invalid appBarIconVisible calculation in Layout docs
Calculation of the `appBarIconVisible` variable in the `Layout` docs for [Hiding icon in AppBar when permanentAt rule is active](https://github.com/react-toolbox/react-toolbox/tree/dev/components/layout#hiding-icon-in-appbar-when-permanentat-rule-is-active) is off by one pixel.

It should be:
```
const appBarIconVisible = this.state.width <= breakpoints[permanentAt];
```
instead of 
```
const appBarIconVisible = this.state.width < breakpoints[permanentAt];
```

Currently the icon disappears while the viewport is 1280px wide even though NavBar is not visible yet (only appears at 1281px). Without the fix the icon is only visible while the viewport is 1279px wide. Changing the comparison to `<=` fixes it. 

Here's the change diff:
```diff
diff --git a/components/layout/readme.md b/components/layout/readme.md
index 3870a28..3ad0e66 100644
--- a/components/layout/readme.md
+++ b/components/layout/readme.md
@@ -264,7 +264,7 @@ class MyComponent extends React.Component {
 
   render() {
 +    const permanentAt = 'lg';
-+    const appBarIconVisible = this.state.width < breakpoints[permanentAt];
++    const appBarIconVisible = this.state.width <= breakpoints[permanentAt];
    
     return (
       <Layout>
```

olegstepura:
Fixes #1231

### DIFF
--- a/components/layout/readme.md
+++ b/components/layout/readme.md
@@ -264,7 +264,7 @@ class MyComponent extends React.Component {
 
   render() {
 +    const permanentAt = 'lg';
-+    const appBarIconVisible = this.state.width < breakpoints[permanentAt];
++    const appBarIconVisible = this.state.width <= breakpoints[permanentAt];
    
     return (
       <Layout>


javivelasco:
Thank you!!

PR #1244, opened 2017-02-14
Export all from table

olegstepura:
Export `TableRow`, `TableHead`, `TableCell` as well as `Table` from './table'


olegstepura:
You cannot write `import { Table, TableRow, TableHead, TableCell } from 'react-toolbox'` atm. Should be fixed with this PR.

javivelasco:
Makes total sense, thanks!

PR #1247, opened 2017-02-14
fix typo in app_bar readme.md

smidaamine:
the example provided in the doc cause warning, its explained in this issue https://github.com/webpack/webpack/issues/2362, to avoid this issue import `import Link from 'react-toolbox/lib/link'` instead of `import Link from 'react-toolbox/lib/Link'`


javivelasco:
Thanks!

PR #1251, opened 2017-02-16
Update README.md for PostCSS instead of SASS

alexhisen:


javivelasco:
Great work! Thank you!! :)

alexhisen:
Thanks. In my opinion, the project is ready for a 2.0 (non-beta) release.

javivelasco:
I think maybe it makes sense to release 2.0 as stable and to call the refactor to 3.0

supirman:
If I use `create-react-app`, do I need to include postcss too instead of using react-toolbox-themr guide?

javivelasco:
Not really! You can use react-toolbox-themr to generate your custom stylesheets and forget about configuration.

olegstepura:
@alexhisen react-toolbox-example is adapted for react-toolbox 2.0, but isn't updated and contains some basiс setup. Your statement `It hasn't yet been updated for 2.0!` is wrong

PR #1255, opened 2017-02-18
Allow to reference Input via Autocomplete

manuquentin:
I have a use case where I've to call the `focus()` method of the `Input` component from `Autocomplete`. And I think others can have to same.

To be able to do it, `Autocomplete` should be referenceable in the `themr` wrapper.


javivelasco:
Makes sense! Thanks :)

PR #1271, opened 2017-02-26
Remove unnecessary reference to sliderLength in knobOffset method

dan-kwiat:
The offset is currently multiplied then divided by sliderLength, and the test description is misleading.


javivelasco:
It's ok to me but it has linter errors!

dan-kwiat:
Sorry, I've fixed it 👍

javivelasco:
Thanks!

PR #1273, opened 2017-02-27
Custom label and value keys for Dropdown

eriksape:
This PR makes it possible to use custom keys for props label and value.

I added the props `labelKey` and `valueKey` for make easy the the use of custom `label` and `value` in the `source`.


javivelasco:
Makes sense. Beware there is a rewrite of this component coming

javivelasco:
Thanks!

PR #1282, opened 2017-03-02
Fix Mobile Safari issues.

izakfilmalter:
1. Toolbox Spec doesn't load on iOS.
image

2. Inputs have a border-raidus value from user-agent style sheet on iOS.
image


izakfilmalter:
@javivelasco I am not sure if I need to do the 'babel-polyfil' for building lib as well.

javivelasco:
I think the polyfill might be needed indeed for building lib as well. Also we can change the implementation to skip unsupported APIs. What do you think?

javivelasco:
Thank you!

PR #1287, opened 2017-03-03
Slider drag callbacks

izakfilmalter:
Add some callbacks to `Slider` for when the user starts dragging, and stops dragging.


dan-kwiat:
Nice, I'd made similar changes locally but not got round to a PR

javivelasco:
Thank you!

PR #1289, opened 2017-03-04
Autocomplete accepting suggestionMatch none

raphaguasta:
I needed to use the autocomplete without the suggestion match, because i was filtering the results in my back end, so the Autocomplete don't need to filter nothing.

I create the value "none" to the option "suggestionMatch", where the autocomplete don't filter the words.


stavarengo:
UP UP UP

raphaguasta:
Can we think about implement it? @javivelasco

javivelasco:
Good workaround. This is going to be reimplemented @raphaguasta, I'm back on it very soon, sorry for being this out. I'm moving to BCN, searching for an apartment... not quitting!

raphaguasta:
Hi @javivelasco! No problems, I mentioned you because I need this feature in a project in my work. Thank you very much! :D

PR #1290, opened 2017-03-04
Add chinese localization

wowlusitong:
Add chinese localization

Based on
https://github.com/moment/moment/blob/develop/locale/zh-cn.js
https://github.com/moment/moment/blob/develop/locale/zh-hk.js
https://github.com/moment/moment/blob/develop/locale/zh-tw.js

image


javivelasco:
Amazing!

PR #1298, opened 2017-03-07
Fix radio-buttons for touch-devices

gerhardsletten:
A toch-event on the parent label should fix this


javivelasco:
Good workaround, thanks!

gerhardsletten:
@javivelasco  Sorry, have later discovered that a "fast-click" plugin caused this not working, should I update the pull-request removing this change, or can you unmerge it yourself

javivelasco:
I can remove it, no problem

PR #1303, opened 2017-03-07
Dev dialog overflow

theskillwithin:
strategies if dialog is longer than the screen height.


jacargentina:
I'm having a hard time trying to use this feature: there is no docs about the `oveflow` theme key. Please, how do i must use it to allow the dialog contents to scroll? Cheers!

Issue #1306, opened 2017-03-08
Should components accept elements instead of strings for labels?


rafarovina:
Hi,

I think you could replace the propType of some labels from PropTypes.string to PropTypes.element.
I have a translation component. I never have "pure strings" so I always get erros similar to this:
> Invalid prop `title` of type `object` supplied to `Dialog`, expected `string`.

Do you agree?
If yes, is there any way I can help? I'm new to open-source but I'd be happy to help =)

Thanks

rubenmoya:
I opened a PR to allow both, `string` and `element` for labels.

javivelasco:
Hi @rafarovina! I'm not 100% agree on this but I understand the issue. [Here](https://github.com/react-toolbox/react-toolbox/issues/1011) you have a similar issue. The solution in this case is to use a helper to decorate your components and to inject a `translate` function that allows you to access the intl API without declaring a component.

The reason why I'm against this is not because it's not useful but because AFAIK it's not possible to change the tagName for the rendered component in most cases. So you might end up having a `<span><span>Whatever text</span></span>`: not cool.

I'm pasting my approach here anyway:

```jsx
import { compose } from 'ramda';
import { PropTypes } from 'react';
import { getContext, mapProps } from 'recompose';

const withTranslate = () => compose(
  getContext({ intl: PropTypes.object }),
  mapProps(({ intl, ...rest }) => ({
    translate: intl.formatMessage,
    ...rest
  }))
);
```

I'm leaving it open for discussion, I think it's an interesting topic.

rafarovina:
@javivelasco I understand your point.
For translation purposes I see that we can have some workarounds.
But for styling, like the "required" asterisk, you'll need a separate span so you can make it red.

About the discussion on #1309 of accepting PropTypes.node or not, I still think it would be useful. Or else, if I had to do something like the image below, I'd need to wrap it in another span.
![TextField](https://cloud.githubusercontent.com/assets/9036143/23757071/860b260a-04c4-11e7-9e57-0dd7b5c0bab9.png)
```javascript
    const label = [
      <span style={{ color: 'red' }}>C</span>,
      <span style={{ color: 'green' }}>o</span>,
      <span style={{ color: 'blue' }}>l</span>,
      <span style={{ color: 'red' }}>o</span>,
      <span style={{ color: 'green' }}>r</span>,
      <span style={{ color: 'blue' }}>s</span>,
    ];
```

zefj:
[...]

PR #1309, opened 2017-03-08
change label propTypes to allow element

rubenmoya:
According to https://github.com/react-toolbox/react-toolbox/issues/1306 when using a React element for the label (because of translations for example) raises an invalid proptype error.

I changed all the label proptypes to allow both, string and element.

### DIFF
--- a/components/button/BrowseButton.js
+++ b/components/button/BrowseButton.js
@@ -19,7 +19,10 @@ const factory = (ripple, FontIcon) => {
         PropTypes.element,
       ]),
       inverse: PropTypes.bool,
-      label: PropTypes.string,
+      label: PropTypes.oneOfType([
+        PropTypes.string,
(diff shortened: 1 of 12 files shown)


rafarovina:
Hi @rubenmoya.
Please check if you can use PropTypes.node for this.
```
// Anything that can be rendered: numbers, strings, elements or an array
// (or fragment) containing these types.
optionalNode: React.PropTypes.node,
```

> [React Proptypes](https://facebook.github.io/react/docs/typechecking-with-proptypes.html)

rubenmoya:
I thought about it @rafarovina but it says you can use arrays of those types, which I don't think would work for a label.

javivelasco:
Thanks for the PR @rubenmoya but I'd like to evolve the discussion in the original PR to decide if this should be done and which components should be changed :)

PR #1317, opened 2017-03-09
feat(slider): add buffer prop to Slider

lucasljj:
I'm making custom controls for html5 videos, and maybe it help someone too


javivelasco:
Thanks!

PR #1319, opened 2017-03-10
Properly register and unregister events on Overlay

codrin-iftimie:
The `.bind` that I've removed returns a new function that won't be unregistered when unmounting the component.
Also I've added the event if the components `nextProps` has `onEscKeyDown` set to `true`.

I've looked over the code and saw the Dropdown component has the same [treatment](https://github.com/react-toolbox/react-toolbox/blob/dev/components/dropdown/Dropdown.js#L86). I've done the same for consistency.

I can go for a simpler approach where I remove only the `bind` issue. 

Let me know which path you prefer.

**P.S**: Wonder-full code in this repo. Very well written.
Cheers

### DIFF
--- a/components/overlay/Overlay.js
+++ b/components/overlay/Overlay.js
@@ -2,6 +2,7 @@ import React, { Component, PropTypes } from 'react';
 import classnames from 'classnames';
 import { themr } from 'react-css-themr';
 import { OVERLAY } from '../identifiers';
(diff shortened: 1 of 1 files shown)


javivelasco:
Sorry! I've just merged a PR fixing the same thing and didn't realized this was there before. Thanks for your work on it, sorry I missed it :(

PR #1324, opened 2017-03-13
Fix DatePicker month alignment (#1221)

### LINKED ISSUE #1221: DatePicker month title is misaligned with next and prev buttons
There are actually two separate issues here:
1) The month title defaults to 3 rem, whereas the buttons are 3.6rem, so they were not perfectly aligned in 1.x and still aren't in 2.0.0. The title, I think, should be 3.6rem.
2) In 2.x, button > span has vertical-align: middle, which makes it really off. It should be vertical-align: top for those next/prev arrow buttons.

Current work-around for both in a theme:
.title {
  line-height: 3.6rem!important;
}

.next > span, .prev > span {
  vertical-align: top!important;
}

rubenmoya:


javivelasco:
Thank you!

PR #1325, opened 2017-03-13
Tabs pointer not being updated (#1224)

### LINKED ISSUE #1224: Pointer in Tabs does not update when props are updated.
I wrapped Tabs inside my own component:

```
return (
      <div className={navBarClass}>
        {!firstLevel && <IconButton />}
        <Tabs
          index={index}
          onChange={this.handleTabChange}
          {...this.props}
        >
          {data.map((item, mapIndex) => {
            return (
              <Tab
                key={mapIndex}
                label={item.label}
                onActive={this.handleActive}
                style={fixedWidth ? tabStyle : undefined}
              />
            );
          })}
        </Tabs>
    </div>
);
```

When my component receives new `props.data` it the display of data is working fine. However, the width for the pointer is not calculated properly and only updates when resizing or clicking another tab.

Before click: [http://imgur.com/a/SpI6A](http://imgur.com/a/SpI6A)

After click: [http://imgur.com/a/4KP0I](http://imgur.com/a/4KP0I)

When new labels are displayed the pointer always keeps the width of the previous label displayed.

Any ideas why this is happening?

rubenmoya:
Fixes #1224.

I also added a Jest config file, to mock `requestAnimationFrame`, I've been thinking about where to place it, some projects place it in `/config/jest` (like create-react-app), jest for example has it in the root folder... so I think for now (since we don't have any other configs) it's ok to place it in the root.

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -51,8 +51,13 @@ const factory = (Tab, TabContent, FontIcon) => {
       this.handleResize();
     }
 
-    componentWillReceiveProps(nextProps) {
-      this.updatePointer(nextProps.index);
+    componentDidUpdate(prevProps) {
+      const { index, children } = this.props;
+      const { index: prevIndex, children: prevChildren } = prevProps;
+
+      if (index !== prevIndex || children !== prevChildren) {
+        this.updatePointer(index);
+      }
     }
 
     componentWillUnmount() {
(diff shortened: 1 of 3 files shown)


javivelasco:
Tests are still failing!

javivelasco:
Thanks!

PR #1326, opened 2017-03-13
Fixes table clone element when child is null (#1272)

### LINKED ISSUE #1272: (2.0.0-beta.6) TableHead and TableRow don't support null or false child nodes
Both TableHead and TableRow produce an error when one of their children is a `null` or `false` value, which makes conditional rendering of child TableCells problematic.

A minimal test case is the following:

```
import React, { Component } from 'react';
import ReactDOM from 'react-dom';
import { Table, TableHead, TableCell } from 'react-toolbox/lib/table';

class Test extends Component {
  render() {
    const { condition } = this.props;
    return (
      <Table>
        <TableHead>
          <TableCell>This renders fine</TableCell>

          {(condition !== false) && (
            <TableCell>This causes an error</TableCell>
          )}
        </TableHead>
      </Table>
    );
  }
}

ReactDOM.render((<Test condition={false} />), document.getElementById('root'));
```

The issue appears to be caused by the use of `cloneElement` as in the following snippet from TableHead.js:50:

```
 {React.Children.map(children, (child, index) => cloneElement(child, {
            column: index,
            tagName: 'th',
  }))}
```

In this case when one of the children is `null` or `false` cloneElement can't handle the input and causes an error.

rubenmoya:
Fixes #1272. Not sure if `if (!child) return null;` is the best (or most beautiful) way to do it.

### DIFF
--- a/components/table/TableHead.js
+++ b/components/table/TableHead.js
@@ -47,10 +47,13 @@ const factory = (Checkbox, TableCell) => {
               onChange={this.handleSelect}
             />}
           </TableCell>}
-          {React.Children.map(children, (child, index) => cloneElement(child, {
-            column: index,
-            tagName: 'th',
-          }))}
+          {React.Children.map(children, (child, index) => {
+            if (!child) return null;
+            return cloneElement(child, {
+              column: index,
+              tagName: 'th',
+            });
+          })}
         </tr>
       );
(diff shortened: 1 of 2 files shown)


javivelasco:
LGTM :shipit:

Issue #649, opened 2016-07-16
Autocomplete issue on shouldComponentUpdate


carlitux:
I am trying to use Autocomplete but for some reason on shouldComponentUpdate not working well. The component is about two levels from root component. But if I set the direction different than 'auto'... it works as should be.

On this line https://github.com/react-toolbox/react-toolbox/blob/dev/components/autocomplete/Autocomplete.js#L70 I added a log:

```js
console.log(this.state.focus, nextState.focus, this.props.direction, POSITION.AUTO)
```

logs
```js
false true "auto" "auto"
```

and this line https://github.com/react-toolbox/react-toolbox/blob/dev/components/autocomplete/Autocomplete.js#L74 looks like doesn't fire the lifecycle **sometimes**, I did a quick test with this code and works (I don't feel is the best approach):

```js
 shouldComponentUpdate (nextProps, nextState) {
   if (!this.state.focus && nextState.focus && this.props.direction === POSITION.AUTO) {
    const direction = this.calculateDirection();
     if (this.state.direction !== direction) {
       // not calling setState
       this.state.direction = direction;
     }
   }
   return true;
 }
```

Reading https://facebook.github.io/react/docs/component-specs.html

docs from shouldComponentUpdate doesn't have a suggest if we can use setState or not...

farwayer:
https://github.com/react-toolbox/react-toolbox/pull/646

javivelasco:
Fixed

PR #1327, opened 2017-03-14
Fix Autocomplete direction calculation

LachlanStuart:
Related issue: https://github.com/react-toolbox/react-toolbox/issues/649

The fix implemented in issue #649 did not appear to fix the behaviour of Autocomplete. Calling `setState` during `shouldComponentUpdate` doesn't result in the new state being set before render is called. This meant that the first render of the Autocomplete would have the previously calculated `direction` value, and this wouldn't be fixed until the next render.

This change moves the calculation of `state.direction` out of `shouldComponentUpdate` so that it is reliably updated prior to rendering after the Autocomplete gets focus.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -80,16 +80,10 @@ const factory = (Chip, Input) => {
          query: this.query(nextProps.value)
        });
      }
-   }
-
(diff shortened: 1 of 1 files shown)


javivelasco:
Cool, thanks!

Issue #755, opened 2016-08-31
Autocomplete allowCreate not working


be4code:
While using `allowCreate={true}` an error is thrown after pressing 'Enter' .

nicomfe:
hey @javivelasco Im having this issue, and the fix doesnt seem to help, 

Im using a multiple value autocomplete, when I insert a new value and press enter the chip doesnt get added. Im not seeing any error on the console either.

be4code:
I figured out it is necessary to add manually entered value to the `source`.

26 wrz 2016 23:46 "Nicolas" <notifications@github.com> napisał(a):

nicomfe:
but whats the point of the allowCreate then? if you have to add manually new values.

be4code:
Imho none :) but that's how it works currently

2016-09-27 0:05 GMT+02:00 Nicolas <notifications@github.com>:

javivelasco:
Ha! Maybe it was not released yet? If allowCreate it's not useful it should be removed 😕

matthax:
Perhaps the Autocomplete should actually have a callback for when a new item is created? That way we can easily implement adding the new value to the source, and avoid altering the input source property...

PR #1334, opened 2017-03-16
Fix allowCreate in Autocomplete when value prop is an object

mikkom:
This change fixes the allowCreate functionality of Autocomplete component when the the value prop is an object. The autocomplete spec in dev branch already exhibits the bug: the array based autocomplete in the top adds new items just fine but same doesn't work with the object-based component below it even though both have the allowCreate prop set.

Related to comments in #755.


javivelasco:
Thanks!

PR #1344, opened 2017-03-21
Fixes Ripple event error in case props.ripple === false

k0r8i:
Ripple component tries to attach event handlers in case ripple is set to false; Causes e.g. ListItem component to fail for
`<ListItem ripple={false} />`


javivelasco:
Thank you all!

PR #1382, opened 2017-03-31
add 'onActiveChange' callback to DatePicker

binocarlos:
this makes external management of the 'active' property much simpler

for example - the onFocus event was cancelling the onClick event for the text box

the onFocus handler did not trigger the 'onClick' property and so there was no way to properly manage the 'active' property externally.

this commit adds a callback for whenever the 'active' state changes internally so the external state (e.g. in redux can be updated also)

the result should be:

 * whenever the internal state changes - the external callback triggers updating external state to the internal value
 * if the external state changes - the new value should be passed in as the 'active' property and this.componentWillReceiveProps will call this.setState

these changes should be backwards compatable with existing code as they don't change existing callback handlers or the internal triggers just adds an extra callback

### DIFF
--- a/components/date_picker/DatePicker.js
+++ b/components/date_picker/DatePicker.js
@@ -36,6 +36,7 @@ const factory = (Input, DatePickerDialog) => {
       minDate: PropTypes.instanceOf(Date),
       name: PropTypes.string,
       okLabel: PropTypes.string,
+      onActiveChange: PropTypes.func,
       onChange: PropTypes.func,
       onClick: PropTypes.func,
       onDismiss: PropTypes.func,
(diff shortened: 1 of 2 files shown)


components/date_picker/DatePicker.js
```diff
@@ -70,40 +71,45 @@ const factory = (Input, DatePickerDialog) => {
       }
     }
 
+    toggleActive = (active) => {
```
javivelasco: Since you are passing an argument, I wouldn't call this method `toggleActive` but `setActive` or something like that. `toggleActive` suggest more like changing a boolean to its opposite.

components/date_picker/DatePicker.js
```diff
@@ -70,40 +71,45 @@ const factory = (Input, DatePickerDialog) => {
...
+      if (this.props.onActiveChange) this.props.onActiveChange(active);
+      this.setState({ active });
```
javivelasco: Since calling `onActiveChange` happens right after the actual state is changed, maybe it makes sense to use the callback parameter of `setState` to perform this action:

```jsx
this.setState({ active }, () => {
  if (this.props.onActiveChange) {
    this.props.onActiveChange(active);
  }
})
```

alexhisen:
What happened to this PR?

PR #1386, opened 2017-04-03
Fix TableCell definition

priecint:
As per source (https://github.com/react-toolbox/react-toolbox/blob/dev/components/table/TableCell.js#L7) `sorted` accepts lower-case strings but the TypeScript definition uses upper-case. So there is either error from TS (`TS2322:Type 'string' is not assignable to type '"ASC" | "DESC"'`) or React (`Warning: Failed prop type: Invalid prop `sorted` of value `ASC` supplied to `ThemedTableCell`, expected one of ["asc","desc"].`).


javivelasco:
Thank you!

PR #1388, opened 2017-04-04
Allow and handle float index in tabs

sanfilippopablo:
This allows tabs to handle index values that are not ints. When the index is a float, the pointer will render somewhere between the tabs corresponding to the nearest index. This is useful for a better integration with libraries like [react-swipeable-views](https://github.com/oliviertassinari/react-swipeable-views).

You can look at the demo in the spec.

Here's a gif with it working with react-swipeable-views in an app I'm working on:
react-toolbox and react-swipeable-views

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -80,19 +80,44 @@ const factory = (Tab, TabContent, FontIcon) => {
     };
 
     updatePointer = (idx) => {
-      if (this.navigationNode && this.navigationNode.children[idx]) {
-        requestAnimationFrame(() => {
-          const nav = this.navigationNode.getBoundingClientRect();
-          const label = this.navigationNode.children[idx].getBoundingClientRect();
-          const scrollLeft = this.navigationNode.scrollLeft;
-          this.setState({
-            pointer: {
-              top: `${nav.height}px`,
-              left: `${(label.left + scrollLeft) - nav.left}px`,
-              width: `${label.width}px`,
-            },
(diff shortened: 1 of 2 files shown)


Kikobeats:
@sanfilippopablo This is awesome! 

Could be possible how to do the react-swipeable-views?

Have sense handle the swipe inside toolbox?

sanfilippopablo:
If you're asking how the integration with react-swipeable-views works, it takes advantage of the `onSwitching` callback prop of `SwipeableViews`, that gets called with the float index while swiping. So the full react-toolbox tabs and react-swipeable-views example would look like this:
```es6
import React from "react";
import SwipeableViews from "react-swipeable-views";
import { Tabs, Tab } from "react-toolbox/lib/tabs";

class App extends React.Component {
  state = {
    index: 0
  };

  setIndex = index => {
    this.setState({ index });
  };

  render() {
    return (
      <div>
        <Tabs index={this.state.index} onChange={this.setIndex}>
          <Tab label="First" />
          <Tab label="Second" />
          <Tab label="Third" />
        </Tabs>
        <SwipeableViews index={this.state.index} onSwitching={this.setIndex}>
          <div style={{ height: 200 }}>
            view n°1
          </div>
          <div style={{ height: 200 }}>
            view n°2
          </div>
          <div style={{ height: 200 }}>
            view n°3
          </div>
        </SwipeableViews>
      </div>
    );
  }
}

export default App;

```

PR #1391, opened 2017-04-04
Add hover effect to raised buttons.

PeterCP:
Raised buttons have no hover effect like flat buttons do, and they should have said effect per the material design guidelines. I added this effect using an ::after pseudo-element whose background changes when the button is hovered.

### DIFF
--- a/components/button/config.css
+++ b/components/button/config.css
@@ -2,23 +2,32 @@
   --button-border-radius: calc(0.2 * var(--unit));
   --button-height: calc(3.6 * var(--unit));
   --button-toggle-font-size: calc(2 * var(--unit));
+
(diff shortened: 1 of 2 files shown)


javivelasco:
Thanks for the PR! Why using a pseudoelement?

PeterCP:
Because raised buttons have a colored background and adding the hover effect directly would mean complicated color manipulations. Using pseudo-elements is, in my opinion, easier, and it allows for a more consistent effect since it uses the same tint-above-the-background implementation that flat buttons use.

javivelasco:
Restest this

javivelasco:
Retest this

PR #1395, opened 2017-04-05
Run builds from last node LTS and stable

Kikobeats:


Kikobeats:
The build fails because you are using `--harmony_proxies` flag that doesn't exist in the last node version because proxy support is native.

See: https://github.com/nodejs/node/issues/8388

Kikobeats:
I removed the flag and looks as expected. 

Do @javivelasco remember why you added it?

PR #1423, opened 2017-04-20
Add support for treeshaking

ankurp:
According to Rollup guide adding `module` field in package.json helps tree shaking aware tools like webpack 2 or rollup import the ES6 modules directly to use the tree shaking feature and reduce the bundle size.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "homepage": "http://www.react-toolbox.com",
   "version": "2.0.0-beta.8",
   "main": "./lib",
+  "module": "./components",
   "author": {
     "email": "javier.velasco86@gmail.com",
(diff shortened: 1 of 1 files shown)


ankurp:
@javivelasco can you take a look as its a small change.

javivelasco:
Thank you and sorry for the late response!!!

tomchentw:
This is definitely a cool feature!

I can certainly see it exists in the current `dev` branch with `package.json` saying its version of `2.0.0-beta.11`. But neither does this tag exists in the git tag of `2.0.0-beta.11` or `2.0.0-beta.12`. Something wrong?

tomchentw:
Can anyone making treeshaking to work with latest `2.0.0-beta.12` version?

PR #1439, opened 2017-04-28
Allow to change FontIcon for Tab by passing it into the factory

ankurp:


javivelasco:
It should be injected in the component index too. Currently we are exposing both factories and created components at the index of each component folder. I think this would break the index

ankurp:
@javivelasco Updated PR as per your comment https://github.com/react-toolbox/react-toolbox/pull/1439/files#diff-229e916e3f842cfaec6234351be8756bR12

javivelasco:
Looks cool! Thanks

PR #1442, opened 2017-05-03
Input component multiline does not auto resize if rows is defined

iraudasoja:
**The problem**
If input component has property `rows` defined as number (PropType is number), the text area height is fixed to it and overflowing text will add scroll bar to it. If the property is defined as a string (e.g. "2"), the text area will auto-resize to lines. The problem with this is, that the prop type is defined as a number and the string value is not allowed there.

**Solution**
Add prop `autoResize` to manage auto-resizing. If its value is `true`, text area increase and decrease its height. Otherwise height is fixed. Default value is `true`

### DIFF
--- a/components/input/Input.d.ts
+++ b/components/input/Input.d.ts
@@ -54,6 +54,10 @@ export interface InputTheme {
 
 export interface InputProps extends ReactToolbox.Props {
   /**
+   * If true, a textarea height is adjusted to rows.
+   */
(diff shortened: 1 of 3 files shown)


javivelasco:
Please correct me it I'm mistaken but you can omit the `rows` property and achieve the same result without an additional property. By default a text area increases its size adjusting to the text height

PR #1448, opened 2017-05-06
Updated dependencies

gvillenave:
Switched the react-addons-css-transition-group package (which is being deprecated) to react-transition-group/CSSTransitionGroup, and updated minimum React version to 15.5.0.


javivelasco:
Thanks!

PR #1454, opened 2017-05-10
Fixes #1452

### LINKED ISSUE #1452: Input documentation example
The use of `bind` in the `render` method is discouraged as it creates and allocates a new bound function on each render which, in this case, means for each keystroke.  In the example usage, instead of:
```
onChange={this.handleChange.bind(this, 'name')} 
```
it would be better to have
```
onChange={this.handleChange} 
```
and then:

```
  handleChange = (value, ev) => {
    this.setState({[ev.target.name]: value});
  };
```
Besides, in `setState` there is no need to merge the previous state with the new properties since `setState` already does so.

Satyam:
Avoid using .bind within the render method of components.

I only applied it to the `Input` component which is the one I originally reported.  Other uses of this usage of bind within render also need to be fixed.  Will do in separate PRs

### DIFF
--- a/components/input/readme.md
+++ b/components/input/readme.md
@@ -9,18 +9,18 @@ import Input from 'react-toolbox/lib/input';
 class InputTest extends React.Component {
   state = { name: '', phone: '', email: '', hint: '' };
 
-  handleChange = (name, value) => {
-    this.setState({ [name]: value });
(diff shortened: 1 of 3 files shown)


javivelasco:
Thanks :)

Issue #1472, opened 2017-05-17
TypeScript error - Datepicker.d.ts


miksansegundo:
In the typescript definition file of Datepicker:
```
locale?: "de" | "no" | "en" | "es" | "af" | "ar" | "be" | "bg" | "bn" | "bo" | "br" | "bs" | "ca" | "gl" | "eu" | "pt" | "it" | "fr" | "ru" | "ua" | "zh-cn" | "zh-hk" | "zh-tw" DatePickerLocale;
```
There is a missing pipe bar '|' before *DatePickerLocale*
```
locale?: "de" | "no" | "en" | "es" | "af" | "ar" | "be" | "bg" | "bn" | "bo" | "br" | "bs" | "ca" | "gl" | "eu" | "pt" | "it" | "fr" | "ru" | "ua" | "zh-cn" | "zh-hk" | "zh-tw" | DatePickerLocale;
```

miksansegundo:
I have found other issue in the same file. The type for the **required** property is not defined.

This is the code to fix the issue:
```
  /**
   * Used in case the datepicker is required.
   */
  required?: boolean;
```

hardchor:
^ this is where you open a PR ;)

miksansegundo:
Pull request done :) #1491

PR #1491, opened 2017-05-24
Added required prop type to DatePicker

miksansegundo:


michaelmyers:
@miksansegundo It looks like you are missing the pipe you initially reported in #1472.

javivelasco:
@michaelmyers I think it's already fixed so this PR finishes fixing both issues. Thanks for noticing and for the fix @miksansegundo

PR #1492, opened 2017-05-24
Added source and model prop types to Table

miksansegundo:


javivelasco:
The model and source are deprecated in dev. This should be done against master for a legacy release

PR #1497, opened 2017-05-25
Fix README on example project description

olegstepura:
Fix it since no reaction to https://github.com/react-toolbox/react-toolbox/pull/1251#issuecomment-302403914


javivelasco:
Thanks!

PR #1499, opened 2017-05-25
remove `theme[mode]` from progress_bar

bbassett:
This is creating an error where `classnames` is trying to add a class value from `theme[mode]`. The issue is that there is no `mode` class in the `progress_bar/theme.css` and therefore, a class `undefined` is being added to the progress_bar. 

We love this library, but we're trying to remove all `undefined` classes from the components as we go.


tobiasandersen:
This would also remove the possibility for people to style the modes differently, right?

javivelasco:
This would break the styles. The issue is that there is no `theme.determinate` classname so when `mode` is `determinate` you get an `undefined` class. The fix is to hardcore `theme.indeterminate` when mode is `indeterminate`.

PR #1501, opened 2017-05-26
Added label to InputTheme interface

mirandamon:
Addresses the error:

TS2322: Type '{ input: any; bar: any; label: any; inputElement: any; }' is not assignable to type 'InputTheme'.
      Object literal may only specify known properties, and 'label' does not exist in type 'InputTheme'.

when adding a theme for label in <Input theme={{label: styles.label}} />


mirandamon:
@javivelasco bump

javivelasco:
Thanks!

PR #1509, opened 2017-05-29
#fix <NavDrawer /> z-index

### LINKED ISSUE #1504: Slider knob draws on top of Dropdown values
Hi guys, 
I found a `z-index` conflict so that whenever a dropdown is expanded over a slider, the knob stays on top of the dropdown. Knob's `z-index` is 200 while dropdown values `z-index` is 100.  Check picture below:

rtbissue

rubenmoya:
Fixes #1504 

Not sure if we should change it in <Drawer />'s theme too.

### DIFF
--- a/components/layout/theme.css
+++ b/components/layout/theme.css
@@ -36,7 +36,7 @@
 
 .sidebarDrawer,
 .navDrawerDrawer {
-  z-index: var(--z-index-high);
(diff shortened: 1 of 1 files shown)


javivelasco:
We should find a different solution. It cannot hold the maximum index because a Dialog can be opened on top of an active NavDrawer

PR #1513, opened 2017-05-30
Feature/tabs a11y

Schlesiger:
Add a11y to tabs.

I did my best with the :focus style but I'm not picky as long as there is some tab navigation indicator.

In the future, `aria-controls` should also be configured, but this will be tricky if tab content is somewhere else on the page.

### DIFF
--- a/components/tabs/Tab.js
+++ b/components/tabs/Tab.js
@@ -65,7 +65,7 @@ const factory = (ripple) => {
       }, className);
 
       return (
-        <div {...other} data-react-toolbox="tab" className={_className} onClick={this.handleClick}>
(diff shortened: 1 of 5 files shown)


miguel-orange:
@Schlesiger Great job. Do you know when this will be merged into react-toolbox:dev?

Schlesiger:
Thanks! I'm not sure when it will be merged. The project seems to be in a summer lull. #1596 shares my sentiments. This is a great project but more maintainers should be recruited.

javivelasco:
Great work :) Sorry about the inactivity, I'm trying to get back on this

PR #1515, opened 2017-05-31
Do not show scrollbar on autocomplete component IE11

Doff3n:
Scrollbars are shown on the autocomplete component in IE11.

It is not possible to use the scrollbar after it appears which confuses my users.

I am not sure IE implementations are wanted, but this in any case disables the scrollbar so that the autocomplete looks more like in chrome.


javivelasco:
Thanks!

PR #1518, opened 2017-06-01
Add variables for the App Bar's font size and weight.

tarnasky:
Add variables for the App Bar's font size and weight and align the defaults to Google's Material Design specs, which say "Across form factors, text that appears in the app bar should use the Title style, Medium 20sp." (see https://material.io/guidelines/style/typography.html).


javivelasco:
Makes sense!

PR #1519, opened 2017-06-02
Change tab style for ripple to work with Tabs

osmelmora:
Changing Tab position to relative and overflow hidden.
Not sure if this should be the responsibility of the ripple HOC.


javivelasco:
IMO it's responsibility of the component that uses the ripple so the fix is ok as long as it has no negative impact on styling

PR #1533, opened 2017-06-12
Add `accept` property to `BrowseButton`

Satyam:
* Corresponds to `accept` attribute on `<input type="file"`
* Allows defining which type of files are to be listed
* Defaults to `*/*` which means all MIME types

Not documented in the original, it remains so.
No tests added, there were none to start with. 
I am successfully using the new property in my code.


javivelasco:
Makes total sense, thanks!

PR #1521, opened 2017-06-02
Issue 1459: Replace onClick handler in Dropdown component to onMouseDown

torpeyp:
Fixes https://github.com/react-toolbox/react-toolbox/issues/1459

Issue #1459, opened 2017-05-13
Dropdown onChange event not firing on source select


danieljodeci:
I'm having a strange issue where the onChange property for the Dropdown component won't trigger any events. I'm using the example from the website and using react-toolbox@2.0.0-beta.8 
```import Dropdown from 'react-toolbox/lib/dropdown';

const countries = [
  { value: 'EN-gb', label: 'England' },
  { value: 'ES-es', label: 'Spain'},
  { value: 'TH-th', label: 'Thailand' },
  { value: 'EN-en', label: 'USA'}
];

export default class DropdownTest extends React.Component {
  state = { value: 'ES-es' };

  handleChange = (value) => {
    console.log('please work:', value);
    this.setState({value: value});
  };

  render () {
    return (
      <Dropdown
        auto
        onChange={this.handleChange}
        source={countries}
        value={this.state.value}
      />
    );
  }
}
```

onBlur works fine. looking deeper it seems the onClick event attached the the 'li' element isn't firing the handleSelect function.

brandonfajardo:
I'm having the same problem, any suggestions?

mbilokonsky:
Can confirm that I'm having the same problem as well. Anyone found a solution?

danieljodeci:
i wasn't able to find a solution. ended up using react-select and customizing the styling to match react-toolbox. annoying but worked out. still am curious as to a solution for this issue.

joelzimmer:
Hey all - we have a fix for this pending that we would love anyone to look at - can someone do that?

joelzimmer:
@javivelasco @soyjavi Hey guys, we'd love if you could take a look at this.

mbilokonsky:
If this works, I'd LOVE a merge and bugfix version bump sooner than later...

tobiasandersen:
Hm, I'm unable to reproduce this. If someone seeing this could create a minimal reproducible example, it would be really helpful. I'm not sure about replacing `onClick` with `onMouseDown` (as proposed in #1521), it feels more like a hack.

Issue #1354, opened 2017-03-24
beta-7: Dropdown onBlur called on focus


ben-freenome:
Dropdown's onBlur function is called on focus, not blur. This seems to be what is leading to the problem where tabbing through multiple dropdowns does not close the previous dropdown, so that all previous dropdowns remain open.

flushentitypacket:
Just ran into this as well. Seems like it'd be as simple as removing this line:
https://github.com/react-toolbox/react-toolbox/blob/3d8cd6689b61760c7b5a39d45c2ed0af3c968403/components/dropdown/Dropdown.js#L102

burtontanner:
I also cannot get the onBlur prop to fire correctly.

Schw2iizer:
I am also having the same issue.

Issue #1323, opened 2017-03-12
Dropdown div's onBlur triggered before <li> onClick


landabaso:
This happens on my App using react 2 beta 6:

The dropdown parent `<div>` onBlur callback is triggered before the `<li>` onClick.
I never get the onChange because the dropdown is closed by onBlur before the onClick is captured.

I'm not doing anything particularly strange with the Dropdown component but there may be some interaction with the particular css/DOM of my App.

It's strange nobody else got problems with that.
May be related with this: [http://stackoverflow.com/questions/17769005/onclick-and-onblur-ordering-issue](url)

I'm in a hurry now to post further details. I will come back to this issue and add more details when I have a bit more time.

No one else having issues with this one?

landabaso:
I checked a bit more.
The parent's onBlur is definitely triggered before the `<li>` onClick.
Dropdown will not work depending on how React does the reconciliation I guess.
If the parent component triggers a render in the middle of the onBlur/onClick then you'll component will render the state active:false before getting the onClick.

landabaso:
Round 3:
**SIMPLE FIX**:
This issue can be fixed using onMouseDown / onTouchStart instead of onClick in the <li> element.
I'm in a hurry now. Could post a PR in a week if nobody else fixes that before. If you can fix it, pleas ego ahead!

matthax:
They still don't close for me. Once the dropdown is opened, if a value is clicked once the focus and select events conflict with each other. I fixed the issue by pausing the event. Here's the relevant code change.

```js
handleSelect = (item, event) => {
  events.pauseEvent(event);
  if (this.props.onBlur) this.props.onBlur(event);
  if (!this.props.disabled && this.props.onChange) {
    if (this.props.name) event.target.name = this.props.name;
    this.props.onChange(item, event);
    this.close();
  }
};
```

VineetJalgaonkar:
I came across this issue. but I am not using the react-toolbox. I created div's. On parent div there is onBlur and on the data div i.e. child div there is onClick. There is specific version of IE 11.0.9.17 there its calling the parent event(onBlur) first when I clicked on the Child div. In the rest of the browsers its working perfectly.

khantopa:
Execute the onBlur in setTimeout  like 

```
handleBlur = () => {
  setTimeout(() => {
   // Things to do onBlur 
  },100)
}
```

tettoffensive:
[...]

PR #1548, opened 2017-06-21
Fix bug where dropdowns don't close

tobiasandersen:
As I'm sure you're aware of, there are a few problems with the Dropdown component. To me, it seem to be mainly these two:

1. Dropdowns don't close when another is opened.
2. In some situations, the `onChange` event isn't fired when clicking on an item.

I know you've been working on a [new implementation](https://github.com/react-toolbox/react-toolbox/issues/1170), but I still feel like we should address these problems in the current implementation as well. I believe all these issues/PRs are related:

- https://github.com/react-toolbox/react-toolbox/issues/1354
- https://github.com/react-toolbox/react-toolbox/issues/1323
- https://github.com/react-toolbox/react-toolbox/pull/1521
- https://github.com/react-toolbox/react-toolbox/issues/1459
- https://github.com/react-toolbox/react-toolbox/issues/1246

[#1521](https://github.com/react-toolbox/react-toolbox/pull/1521) solves part of the problem by listening for `mousedown` rather than `click` on the dropdown items.  I don't really like this solution since it feels like a hack that could bite us further ahead (`click` and `mousedown`  really isn't the same). 

For the problem where dropdowns don't close, I believe it stems from [this line](https://github.com/react-toolbox/react-toolbox/blob/dev/components/dropdown/Dropdown.js#L133): 
`if (this.inputNode) this.inputNode.blur();`
Which is called whenever a dropdown is opened. I'm not exactly sure why it was added, but I can see it had something to do with [issues on iOS](https://github.com/react-toolbox/react-toolbox/issues/866). The reason it's problematic is that the component doesn't ever have "real" focus, and thus the `blur` event won't fire when another dropdown is opened. 

This PR tries to work around this by: 
1. Removing the parts which call `blur` programmatically on the input node.
2. Allow the dropdown to have focus by giving its container div `tabIndex="-1"`. This also eliminates the need for the hack in #1521.

It'd be great if others could try this as well, since I've only done some really basic testing myself (Chrome only). And I don't have access to a device running iOS, so I can't tell if this is messing things up over there.

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -130,7 +130,6 @@ const factory = (Input) => {
       const client = event.target.getBoundingClientRect();
       const screenHeight = window.innerHeight || document.documentElement.offsetHeight;
       const up = this.props.auto ? client.top > ((screenHeight / 2) + client.height) : false;
-      if (this.inputNode) this.inputNode.blur();
       this.setState({ active: true, up });
     };
 
@@ -207,6 +206,7 @@ const factory = (Input) => {
           data-react-toolbox="dropdown"
           onBlur={this.handleBlur}
           onFocus={this.handleFocus}
+          tabIndex="-1"
         >
           <Input
             {...others}
@@ -215,7 +215,6 @@ const factory = (Input) => {
             onClick={this.handleClick}
             required={this.props.required}
             readOnly
-            ref={(node) => { this.inputNode = node && node.getWrappedInstance(); }}
             type={template && selected ? 'hidden' : null}
             theme={theme}
             themeNamespace="input"


javivelasco:
I think it's done in a different PR, I didn't notice this one, sorry :(

smithgeek:
This change is causing problems where dropdowns are not not closing. It seems to be related to setting `tabIndex="-1"` on the parent div.

We are seeing when you select an item in the list after `close()` is called the input receives focus which causes `open()` to immediately be called which means the dropdown won't actually close.

Just removing the `tabIndex` from the parent div seems to work for me (but I have not tested that much)

PR #1549, opened 2017-06-21
Fix compatibility with TypeScript@2.3.2

freud:
Embedding anything in <Layout> is not possible with TypeScript@2.3.2

### DIFF
--- a/components/layout/Layout.d.ts
+++ b/components/layout/Layout.d.ts
@@ -52,10 +52,6 @@ export interface LayoutTheme {
 
 export interface LayoutProps extends ReactToolbox.Props {
   /**
-   * Children to pass through the component.
-   */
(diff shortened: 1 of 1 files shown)


Pajn:
I'm sorry for not answering in the issue, I thought I did that but apparently I forgot.

I would suggest changing the type to
```ts
  /**
   * Children to pass through the component.
   */
  children?: JSX.Element | JSX.Element[]
```
instead.

If you remove the property older versions of TS wont support a manually passed children prop like this:
```jsx
<Layout children={<Panel />} />
```
Also future versions might have problems with passing children at all (I don't know this, but as the types on children seems to be getting stronger that seems like something that could happen at one point).

The reason for the union type with and without array brackets is to support both a single and multiple elements.

javivelasco:
Closed in favor of #1575

PR #1553, opened 2017-06-23
Add role of input field as a non-required property

jonasknutsen:
The role of the input field is now input. This is an abstract ARIA role, and should be replaced with a non-abstract role. Today the input field will fail an accessibility audit. https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_01

I've made the role a non-required property for the input field, but with default role still as input. This way the role can be changed if desired.


javivelasco:
Cool! Thanks

Issue #1556, opened 2017-06-28
BrowseButton onChange fires twice


nicolazj:
version: 2.0.0-beta.12
was using react-toolbox with `create-react-app`

`import BrowseButton from 'react-toolbox/lib/button/BrowseButton.js';`
`<BrowseButton icon="add" flat primary label="Add files" onChange={_ => console.trace()} />`

the second one is invoked by `ReactErrorUtils`, not sure what is that..

<img width="574" alt="screen shot 2017-06-28 at 3 49 24 pm" src="https://user-images.githubusercontent.com/1446380/27619892-d638ecbe-5c19-11e7-86ee-6599573a4c80.png">

nicolazj:
the root cause seems to be wrapping `input` in `label` tag

nicolazj:
adding `onChange: null,` after https://github.com/react-toolbox/react-toolbox/blob/dev/components/button/BrowseButton.js#L115 can fix the problem

PR #1557, opened 2017-06-28
fixed browserbutton fires onChange event  twice

nicolazj:
fixed browsebutton fires onChange event  twice


nicolazj:
https://github.com/react-toolbox/react-toolbox/issues/1556

javivelasco:
Thanks!

PR #1567, opened 2017-07-03
This should fix bug in Tabs (react-toolbox@1.3.4 ) with incorrect pointer update

timurgarif:
Hi! 
I use master branch (react-toolbox@1.3.4). The bug affects that version.
### Bug description.
When the `Tabs` component contains some `Tab` components that vary depending on an app state, pointer element (under active tab) could be rendered with incorrect width. 
_This doesn't affect initial render, only render after state changes that affect `Tab` components._
E.g. if an active tab receives a new `label` prop that has longer or shorter text than current `label` then `Tabs.updatePointer`  in `Tabs.componentWillReceiveProps` will calculate width at `nextProps.index` (which is correct), **but based on old tab element (which is incorrect)**.
the first state example: first-state
the second state example: second-state
Notice how the blue pointer on the second example repeats the width of the pointer from the first example (previous tabs state).

### Suggested solution.
1. In `Tabs` component remove `componentWillReceiveProps` method.
2. In `Tabs` component add componentDidUpdate method as shown in the pull request.

### dev branch (2.x)
As I noticed, `componentDidUpdate` has been added like in my suggestion. That's nice but I think `componentWillReceiveProps` should be removed there. Though I cannot now test and prove for this version, it most probably triggers unnecessary  render for the old tabs and pointer that will be rerendered by state change in `componentDidUpdate` anyway.

### DIFF
--- a/components/tabs/Tabs.js
+++ b/components/tabs/Tabs.js
@@ -47,8 +47,12 @@ const factory = (Tab, TabContent, FontIcon) => {
       this.handleResize();
     }
 
-    componentWillReceiveProps (nextProps) {
-      this.updatePointer(nextProps.index);
+    componentDidUpdate (prevProps) {
+      const { index, children } = this.props;
+      const { index: prevIndex, children: prevChildren } = prevProps;
+      if (index !== prevIndex || children !== prevChildren) {
+        this.updatePointer(index);
+      }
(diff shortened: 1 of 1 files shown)


javivelasco:
Thanks!

Issue #1443, opened 2017-05-03
Table doesn't recognise TableRow child rendered as a separate component


PSilling:
For example, this doesn't do anything when I try to return cells array into a Table component. Only it's TableHead part get's visibly rendered, while no TableRow is created. Also, this runs without errors:
`var cells = [];
cells.push(<Row />);`

where Row is:
`class Row extends Component {
  render() {
    return(
      <TableRow><TableCell>Ahoj</TableCell></TableRow>
    );
  }
}`

However, this works normally:
`cells.push(<TableRow><TableCell>Ahoj</TableCell></TableRow>);`

tobiasandersen:
That's because the [Table makes sure its rows have the correct typ](https://github.com/react-toolbox/react-toolbox/blob/dev/components/table/Table.js#L13), if not, they won't be rendered. You could create your own Row component, and pass that to the Table factory, basically the same as [this](https://github.com/react-toolbox/react-toolbox/blob/dev/components/table/Table.js#L105).

PSilling:
Thank you for the help. So far I've managed to do this:
`import tableFactory from 'react-toolbox/lib/table/Table.js';
const Table = tableFactory(TableHead, Row);`


But if I do this and try to use <Table> later, I recieve an error: Cannot call a class as a function. So I've tried to use `const Table = new tableFactory(TableHead, Row);` but then I get this error: Cannot read property 'themr' of undefined.


Sorry for the trouble, I'm kinda new to JS so help would be much appreciated.

tobiasandersen:
`const Table = tableFactory(TableHead, Row);` is the correct way to do it. Are `TableHead` and `Row` valid components in your example?

What is the behavior you want to change?

baahrens:
Having the same issue. If I only want to use my own component for rows (which renders react-toolboxs ```TableRow``` as a root, so should be valid) and just the standard ```TableHead```, I should call the tableFactory with ```TableHead```and ```MyRowComponent```, right?

```const Table = tableFactory(TableHead, MyRowComponent)```

Because this also throws the 'Cannot call a class as a function' error. 
What am I missing? Thanks in advance

**Comment by PSilling** [comment]
@tobiasandersen I've tried a lot of things (what @baahrens tried included), but still get the 'Cannot call a class as a function' error.

Right now I'm sitting on this syntax, which is probably wrong doe:
```
const Table = tableFactory(UniversityHead, UniversityRow);

class ContentTable extends Component {
  render() {
[...]

PR #1569, opened 2017-07-05
Add workaround to isComponentOfType for react-hot-loader@^3

tomchentw:
As per discussions in https://github.com/react-toolbox/react-toolbox/issues/1443#issuecomment-312201476 #1148 and #1164, `react-hot-loader` also documents a [workaround][workaround]:

```js
const ComponentType = (<Component />).type;
const element = <Component />;
console.log(element.type === ComponentType); // true
```

Would you consider this version of `defaultChecker`?

[workaround]: https://github.com/gaearon/react-hot-loader/blob/master/docs/Known%20Limitations.md#checking-element-types

### DIFF
--- a/components/utils/is-component-of-type.js
+++ b/components/utils/is-component-of-type.js
@@ -1,3 +1,5 @@
+import React from 'react';
+
 let customChecker;
 
 /**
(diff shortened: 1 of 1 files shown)


javivelasco:
That's ok for me!

tomchentw:
Hooray!

Issue #1558, opened 2017-06-28
ListItem: Button on rightActions ugly when has long text


jacargentina:
On the website playground, change the first <ListItem> from:
```jsx
        <ListItem
          avatar='https://dl.dropboxusercontent.com/u/2247264/assets/m.jpg'
          caption='Dr. Manhattan'
          legend="Jonathan 'Jon' Osterman"
          rightIcon='star'
        />
```
to:

```jsx
        <ListItem
          avatar='https://dl.dropboxusercontent.com/u/2247264/assets/m.jpg'
          caption='Dr. Manhattan'
          legend="Jonathan 'Jon' Osterman"
          rightActions={[<Button label="NEW QUERY ON THIS ITEM" primary raised />]}
        />
```
When using some long text on the _label_ it looks without padding on left and right sides.

![ugly_button_inside_listitem](https://user-images.githubusercontent.com/517228/27648878-63d971b0-5c06-11e7-96d7-6d961be2dc5a.png)

rubenmoya:
I'll have a look at it.

PR #1571, opened 2017-07-05
Apply padding 0 to everything but buttons in ListItemAction

rubenmoya:
Related to: https://github.com/react-toolbox/react-toolbox/issues/1558

The `.itemAction > *` rule was overriding the button padding, since I don't know if it would affect to other elements I just used `:not(button)`, not sure if it's the best way to fix it.

Feedback is welcome!!

screen shot 2017-07-05 at 08 15 33

### DIFF
--- a/components/list/theme.css
+++ b/components/list/theme.css
@@ -116,7 +116,7 @@
   display: flex;
   margin: var(--list-item-child-margin) var(--list-horizontal-padding) var(--list-item-child-margin) 0;
 
-  & > * {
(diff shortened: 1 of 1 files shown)


Pajn:
I feel that it might be better to only target what should be targeted. What I can think of right now is IconButton. Currently it may cause similar issues to many component (for example user provided ones).

javivelasco:
Ok as a temporary solution

Issue #1542, opened 2017-06-18
<Layout> doesn't work with typescript higher than 2.2.2


freud:
Hi,

I get from `webpack` the following error when I use typescript@2.3.2:
```
TS2322: Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Layout> & Readonly<{ children?: ReactNode; }> & Re...'.
  Type '{ children: Element; }' is not assignable to type 'Readonly<LayoutProps>'.
    Types of property 'children' are incompatible.
      Type 'Element' is not assignable to type '[NavDrawer | Panel | Sidebar]'.
        Property '0' is missing in type 'Element'.
```

Here is my code:

```
import * as React from "react";
import { AppBar } from 'react-toolbox/lib/app_bar';
import { Layout, Panel } from 'react-toolbox/lib/layout';

export default class App extends React.Component<any, any> {    
    constructor() {
        super();     
    }
    render() {
        return (
            <Layout>
                <Panel>
                    <AppBar title='Title' />
                    <section>
                        Some content
                    </section>
                </Panel>
            </Layout>
        );
    }
}
```

It works with typescript@2.2.2.
Webpack doesn't return any error and the app works, but VSCode shows the same error as for typescript@2.3.2:
![image](https://user-images.githubusercontent.com/37286/27264432-7c465bee-547e-11e7-90bc-6dc68233842e.png)

When I use typescript@2.5.0-dev.20170618 then I get even more errors:
```
ERROR in [at-loader] ./node_modules/react-toolbox/lib/layout/NavDrawer.d.ts:15:18
    TS2430: Interface 'NavDrawerProps' incorrectly extends interface 'DrawerProps'.
  Types of property 'theme' are incompatible.
    Type 'NavDrawerTheme' has no properties in common with type 'DrawerTheme'.

ERROR in [at-loader] ./node_modules/react-toolbox/lib/layout/Sidebar.d.ts:15:18
    TS2430: Interface 'SidebarProps' incorrectly extends interface 'DrawerProps'.
  Types of property 'theme' are incompatible.
    Type 'SidebarTheme' has no properties in common with type 'DrawerTheme'.

ERROR in [at-loader] ./src/components/App.tsx:22:13
    TS2322: Type '{ children: Element; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Layout> & Readonly<{ children?: ReactNode; }> & Re...'.
[...]

PR #1572, opened 2017-07-05
Widen typing for Layout children to 'any'

jasonvasquez:
Proposed fix for #1542 .  See https://github.com/Microsoft/TypeScript/issues/13618 for additional details as well, TS doesn't appear to have a good solve for this issue at this time.  For now, this change widens the requirement for `Layout` children to be an `any`.

It's not ideal, but does allow react-toolbox to be used with TS 2.3 and 2.4, where this did not work previously.

EDIT:  Targeting `agnostic-typescript`, as this appears to be the current branch npm releases are coming from, but I could be mistaken.  Let me know and I can re-target.


javivelasco:
Actually I'm getting agnostic-typescript branch as a mirror of dev plus some experiences that should be eventually merged so I'd rather make the PR to dev if possible :)

PR #1573, opened 2017-07-05
Added innerRef prop to Input

jasonvasquez:
This change exposes the `innerRef` prop to the `Input` component props, so that users may implement functionality that refers directly to DOM input nodes, esp. in the case of wrapped HOC Inputs from react-css-themr.

P.S.: Targeting agnostic-typescript, as this appears to be the current branch npm releases are coming from, but I could be mistaken. Let me know and I can re-target.


jasonvasquez:
(Tagging #1573 for reference)

PR #1577, opened 2017-07-06
Add default export TypeScript type to IconButton

Pajn:
This is required when using IconButton with react-css-themr


javivelasco:
Thanks!

PR #1586, opened 2017-07-14
Fix date-picker animation in IE11

AlexPrinz:
**Bug description:**

In IE11 the animation for changing the month in the date-picker is not smooth. 

image

This pull-request should fix this problem.


javivelasco:
THANKS!

Issue #1444, opened 2017-05-04
Module build failed: Error: composition is only allowed when selector is single :local class name not in ".raised", ".raised" is weird


burakuluu:
Hi,

I tried to use react-toolbox but this error keeps showing after all the solutions that I've found. I finally decided to ask here.

Error is;

> Module build failed: Error: composition is only allowed when selector is single :local class name not in ".raised", ".raised" is weird

My **postcss.config.js** file looks like;
```
module.exports = {
  plugins: [
    require('postcss-smart-import')({ /* ...options */ }),
    require('precss')({ /* ...options */ }),
    require('autoprefixer')({ /* ...options */ })
  ]
}
```

In my .js file, I imported some react-toolbox modules;
```
import AppBar from 'react-toolbox/lib/app_bar';
import Navigation from 'react-toolbox/lib/navigation';
```

And also my webpack.config.dev.js and webpack.config.prod.js look like;
```
var webpack = require('webpack');
var loaders = require('./webpack.loaders');
var path = require('path');
var CopyWebpackPlugin = require('copy-webpack-plugin');
var HtmlWebpackPlugin = require('html-webpack-plugin');

module.exports = {
  context: path.join(__dirname, "src"),
  devtool: 'inline-sourcemap',
  entry: {
    app: "./app.js",
    vendor: [
      'axios',
      'react',
      'react-dom',
      'react-grid-layout',
      'react-bootstrap',
      'react-redux',
      'react-router',
      'redux',
      'redux-logger',
      'redux-thunk'
    ]
  },
  resolve: {
    root: path.resolve(__dirname),
    alias: {
      'dashboard-app/config': 'src/config',
      'dashboard-app/utils': 'src/utils',
      'dashboard-app/auth': 'src/modules/auth',
      'dashboard-app/dashboard-view': 'src/modules/dashboard-view',
      'dashboard-app/header': 'src/modules/header',
      'dashboard-app/widgets-container': 'src/modules/widgets-container',
      'dashboard-app/assets': 'src/assets',
    },
    extensions: ['', '.js', '.jsx']
  },
  module: {
    loaders: [
      {
        test: /\.jsx?$/,
        exclude: /(node_modules|bower_components)/,
        loader: ['babel-loader'],
        query: {
          presets: ['react', 'es2015', 'stage-0'],
          plugins: ['react-html-attrs', 'transform-class-properties', 'transform-decorators-legacy'],
        }
      },
      {
        test: /\.css$/,
[...]

Issue #1359, opened 2017-03-25
Error: composition is only allowed when selector is single :local


eldyvoon:
ERROR in ./~/css-loader!./~/postcss-loader!./~/react-toolbox/lib/button/theme.css
Module build failed: Error: composition is only allowed when selector is single :local class name not in ".raised", ".raised" is weird

If you need my webpack.config.js it's here : http://pastebin.com/33C1quLS

gvillenave:
I'm seeing the same issue when trying to import { Layout, NavDrawer, Panel } from 'react-toolbox/lib/layout'

sanfilippopablo:
I'm having this issue while using react-toolbox through react-toolbox-themr.

sanfilippopablo:
@eldyvoon In your webpack.config.js you're not enabling CSS modules  for css-loader. Check here: https://github.com/react-toolbox/react-toolbox#setting-up-postcsscssnext-in-webpack

PR #1587, opened 2017-07-16
solve #1444 and #1359.

the1900:
I use webpack version 2.6 and new configure syntax rather than 1.x.

When I follow the instruction on `readme.md`. It always failed on *Basic Usage* with the message below.
``` bash
ERROR in ./~/css-loader?{"modules":false,"importLoaders":1,"localIdentName":"[name]--[local]--[hash:base64:8]"}!./~/postcss-loader/lib!./~/react-toolbox/lib/button/theme.css
  Module build failed: Error: composition is only allowed when selector is single :local class name not in ".raised", ".raised" is weird
```
So I figured out what's wrong. The Default value of `css-loader` is false on new versions. 
It will solve #1444 and #1359.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,6 +58,26 @@ Declare plugins to be used by postcss (as part of webpack's config object):
   },
 ```
 
+Configure webpack 2.x or 3.x loader for .css files to use postcss:
+```js
+      {
+        test: /\.css$/,
+        use: [
+          "style-loader",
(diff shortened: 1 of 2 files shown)


rubenmoya:
+1

the1900:
the additional commit is a bugfix. see commit message.

PR #1598, opened 2017-07-24
Replaced tab label with button element for keyboard accessibility. (v1.3.4)

samsch:
Using an actual button element for the tab labels makes them selectable by keyboard, and is more semantically correct (a11y!).

### DIFF
--- a/components/tabs/Tab.js
+++ b/components/tabs/Tab.js
@@ -59,7 +59,7 @@ class Tab extends Component {
     }, className);
 
     return (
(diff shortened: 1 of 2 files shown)


samsch:
Don't merge this as-is yet. The buttons need to have `type="button"` added to them.

javivelasco:
Ping back when it's ready!

samsch:
@javivelasco Ok, this is updated and ready to merge.

The issue was that without `type="button"` on the tab buttons, if the tabs component was nested in a form, the tab buttons/labels would submit the form.

PR #1604, opened 2017-07-25
requestAnimationFrame will only trigger if the browser window is visable

binhqx:
If the browser tab is put to the background, requestAnimationFrame will trigger when the tab returned to the foreground. When the Tabs Component is removed from the DOM when in the background, the requestAnimationFrame must be canceled to prevent attempting to use a unmounted Component.


javivelasco:
That's a great point! Thank you!

PR #1606, opened 2017-07-26
Use prettier

Graham42:
*The Why:* **Having highly deterministic auto-formatting is life changingly awesome.**

I've configured prettier to use the same width, comma, and quote options
as were being used from the airbnb config. However some of these don't
map exactly so this includes a large formatting commit.

This was mostly automatic. The only catch was a couple places where we use comments to disable eslint rules and those comments got wrapped to a new line. In those few cases, I move the comment to the line above and changed it from `eslint-disable-line` to `eslint-disable-next-line`.

The formatting commit can be reproduced with:
```sh
npm run lint:js -- --fix 
# Then manually fix the 5 or 6 places as described above. 
# Running this command again now should fix all the errors
npm run lint:js -- --fix

#Use the command from .lintstagedrc to format the TypeScript files
prettier --write --print-width 100 --trailing-comma es5 --single-quote '**/*.ts' 
```


Graham42:
Aging out

PR #1617, opened 2017-08-03
Fix incompatibility of TimePicker with Preact

louisremi:
I managed to get react-toolbox to work with Preact (which was mostly a matter of customizing the Webpack config).
I've got one remaining problem that is due to some internal differences between React and Preact VDom libraries. From what I've gathered, components should avoid using node references is componentDidMount or componentDidUpdate without wrapping them in a `setImmediate` or `setTimeout(…, 0)` (see https://github.com/developit/preact/issues/477 ).

With the current logic of componentDidUpdate in TimePickerDialog.js, the component throws a `TypeError: Cannot read property 'handleCalculateShape' of undefined` and interaction is impossible:

Fixing this in Preact seems like a difficult if not impossible task, however fixing it in the TimePicker component is only a matter of wrapping the call to `this.clockNode.handleCalculateShape`in an arrow function so that this.clockNode is not accessed immediately on component update.

Let me know if the change is acceptable :-)


Pajn:
I've seen that error in React too in specific cases but never researched it as it was simple enough to work around, this might fix that to and isn't directly a downside anyway

matthax:
I'm running into this issue when using the `Create React App` and with the timepicker inside a `Tab` component...

Pajn:
I think I've realized when the error occurs. It's when you directly render the TimePicker with active=true. Without this PR, you must first render it with active=false and then set it to true.

matthax:
@Pajn I was actually able to work around the issue using something like this:
```jsx
 { change_time ? <TimePicker active label='Real Timestamp' format='ampm' onChange={(value) => { this.handleChange('time', value); this.setState({change_time: false}); }} value={time} /> : <Input type='text' label='Timestamp' readOnly value={format_time(time)} onClick={(e) => { this.setState({ change_time: true }); }} />}
```

I suppose it would be cleaner to do something with `componentDidMount` and a timeout/animation frame but I'm just testing so whatever. 

With that said, I'd like to see a fix in just for the sake of being able to say "here's a working component library, start with this" - and not have to worry about rendering issues lol

PR #1622, opened 2017-08-08
Add option to round time picker to specific intervals.

smithgeek:
Added an option to the time picker component that allows you to round the selector to a given interval (for example only allow selecting in 15 minute increments).

### DIFF
--- a/components/time_picker/TimePicker.d.ts
+++ b/components/time_picker/TimePicker.d.ts
@@ -173,6 +173,10 @@ export interface TimePickerProps {
    * Additional attributes passed to inner Input component.
    */
   [key: string]: any;
+  /**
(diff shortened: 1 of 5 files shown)


Pajn:
I assume that the `roundTo` prop should be set to a number of minutes, for example 5 to allow selecting every 5 minute mark. Maybe add a short sentence in the doc for that to make it crystal clear?

Also please update the readme file as well which is where the homepage is built from.

louisremi:
Does the rounding occur during or after interaction? If it occurs during interaction, the proper name for the option would be "step", I believe.

smithgeek:
It occurs during interaction, i'll change the name to step.

PR #1624, opened 2017-08-09
Fix lint issue

Pajn:
As the lint fails on dev Travis marks the build as failing for all new PRs, this fixes that.


javivelasco:
Tanks @Pajn !

PR #1627, opened 2017-08-09
Disabled input should be dashed not dotted

olegstepura:
According to spec example 
https://material.io/guidelines/components/text-fields.html#text-fields-states

2017-08-09 12_56_19-text fields - components - material design guidelines

### DIFF
--- a/components/input/theme.css
+++ b/components/input/theme.css
@@ -158,7 +158,7 @@
 }
 
 .disabled > .inputElement {
-  border-bottom-style: dotted;
+  border-bottom-style: dashed;
   color: var(--input-text-disabled-text-color);
 }
 


rubenmoya:
In the spec it says `dotted`, although it's true it looks like dashed

screen shot 2018-01-24 at 18 04 50

olegstepura:
@rubenmoya  Dotted is indistinguishable from solid without closer inspection. This was the reason I noticed this and created a PR. I suggest to re-review this.

rubenmoya:
I've just checked in on the playground, you're totally right, it's almost indistinguishable. Let's go for dashed :)

Thank you!

PR #1632, opened 2017-08-09
Fix travis build broken in #1515

olegstepura:
Currently build fails due to an issue introduced in #1515.  Maybe more will follow. For me pre-commit hook is broken on windows (#1631), had to skip it.


olegstepura:
Sorry, did not notice #1624

javivelasco:
Thanks @olegstepura !

PR #1649, opened 2017-08-15
Add helper text to input

olegstepura:
Add ability to insert Helper text
https://material.io/guidelines/components/text-fields.html#text-fields-layout

image

### DIFF
--- a/components/input/config.css
+++ b/components/input/config.css
@@ -12,6 +12,7 @@
   --input-text-disabled-color: var(--input-text-bottom-border-color);
   --input-text-disabled-text-color: var(--input-text-label-color);
   --input-text-error-color: rgb(222, 50, 38);
+  --input-text-helper-color: color(var(--color-black) a(26%));
   --input-text-required-color: rgb(222, 50, 38);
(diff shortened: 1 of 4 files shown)


olegstepura:
@javivelasco Do we all need to modify our PRs merging current master so that travis stops failing to build? Or will you merge it as is?

components/input/config.css
```diff
@@ -12,6 +12,7 @@
   --input-text-disabled-color: var(--input-text-bottom-border-color);
   --input-text-disabled-text-color: var(--input-text-label-color);
   --input-text-error-color: rgb(222, 50, 38);
+  --input-text-helper-color: color(var(--color-black) a(26%));
```
rubenmoya: Could we change this to `--color-text` so it's consistent with the changes made in this PR? https://github.com/react-toolbox/react-toolbox/pull/1766

olegstepura: Well, it has to be lighter than the regular text. If you wish to replace `--color-black` with `--color-text` - that's ok!

Issue #1659, opened 2017-08-25
TimePicker selecting time broken when body scrolled 


peterdivvito:
When the time picker is shown on a page that is currently scrolled, selecting a time is buggy. Diagnosing the problem it seems that the clock face, as far as touching the screen goes, is offset from the actual display of the clock face. 

I tracked the problem down to code in Clock.js

```javascript
handleCalculateShape = () => {
    const { top, left, width } = this.refs.placeholder.getBoundingClientRect();
    this.setState({
      center: { x: left + width / 2 - window.pageXOffset, y: top + width / 2 - window.pageXOffset },
      radius: width / 2
    });
  };
```
Determining the 'center' is taking into account the scroll position of the page.

'center' is then used in ClockHand.js

```javascript
touchStart (event) {
    events.addEventsToDocument(this.getTouchEventMap());
    this.move(events.getTouchPosition(event));
    events.pauseEvent(event);
  }
```
where getTouchPosition is
 
```javascript
getTouchPosition (event) {
    return {
      x: event.touches[0].pageX - (window.scrollX || window.pageXOffset),
      y: event.touches[0].pageY - (window.scrollY || window.pageYOffset)
    };
  },
```

The issue arrises because 'center' calculated above is in body/page coordinates BUT the X & Y returned by getTouchPosition is in viewport coordinates. So calculating the touch position radius (i.e the touch position relative to 'center') will result in unexpected results.

PR #1660, opened 2017-08-25
Issue #1659 time picker touch fix

peterdivvito:


Pajn:
Hmm, the scroll adjustment was explicitly added https://github.com/divvito/react-toolbox/commit/0d1265e5a1a0785434c9e919c2476b64e1532a3e

As you just dug in there, may you see why it would have been added as such? For example, does simply removing the calculation still works in all cases (touch, mouse, scroll on different elements is the immediate things I can think of)?

peterdivvito:
My guess is that the commit made changes to take into account scrolling but went too far.

The changes made to utils/events.js was to turn page level coordinates (pageX & pageY) into viewport coordinates. 

IMHO this would have been enough to fix the issue as the coordinates calculated in handleCalculateShape in Clock.jsx were already in viewport coordinates as that's what getBoundingClientRec() returns (https://developer.mozilla.org/en/docs/Web/API/Element/getBoundingClientRect).

Subtracting the scroll from these coordinates was a step too far.

PR #1664, opened 2017-08-30
Portal: Fix Invalid "style" PropType

adamhenson:
`PropTypes.style` is invalid. In certain cases - seeing this warning in a dev environment.

```
Warning: Failed prop type: Portal: prop type `style` is invalid; it must be a function, usually from React.PropTypes.
    in Portal
```

### DIFF
--- a/components/hoc/Portal.js
+++ b/components/hoc/Portal.js
@@ -1,13 +1,14 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import ReactDOM from 'react-dom';
+import styleShape from 'react-style-proptype';
 
 class Portal extends Component {
   static propTypes = {
     children: PropTypes.node,
     className: PropTypes.string,
     container: PropTypes.node,
-    style: PropTypes.string,
+    style: styleShape,
   }
 
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


Pajn:
Thank you, however string is invalid as style should be an object in React. Could you change this to PropTypes.object ?

adamhenson:
Thanks @Pajn. Looks like we use [`react-style-proptype` in `Check`](https://github.com/react-toolbox/react-toolbox/blob/ae09770cd4aab861e95a070b975d7fdef1ace082/components/checkbox/Check.js#L4). Updated in 2ef22a1.

Pajn:
Good catch!

adamhenson:
No prob @Pajn. You'll notice CI fails. I have a PR to fix that too (#1665 ).

Pajn:
Yeah, the CI situation is not optimal as of now but as you noticed it's not your fault.
There are two previous PR open for that as well #1632 and #1624. Javi has not been able to be very active on this project lately due to personal life. It's not dead, he have plans for its future but that is something that occasionally happens in projects that are not sponsored by a company. All PRs are still greatly appreciated and most should be mergeable when time allows.

adamhenson:
No prob @Pajn - I will close my other PR then. Yes, it's a great project.

Issue #1602, opened 2017-07-24
Migrate to React-Transition-Group v2 - Use Transition instead of CSSTransitionGroup


zxlin:
v2 of React-Transition-Group removed `CSSTransitionGroup` and instead should use just `Transition`

This is fairly painless, just a few css class name changes and replacing the component.

[React-Transition-Group v1 Migration Guide](https://github.com/reactjs/react-transition-group/blob/master/Migration.md)

PR #1701, opened 2017-09-30
Upgrade react-transition-group

Merbs:
Resolves issue https://github.com/react-toolbox/react-toolbox/issues/1602
sep-29-2017 17-36-03
sep-29-2017 17-36-14

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "ramda": "^0.23.0",
     "react-css-themr": "^2.1.2",
     "react-style-proptype": "^3.0.0",
-    "react-transition-group": "^1.1.3"
+    "react-transition-group": "^2.2.1"
   },
   "devDependencies": {
     "@types/node": "^7.0.4",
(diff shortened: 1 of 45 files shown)


andrefox333:
This is great and we definitely could use this in our project since we have another package that depends on version 2 of react-transition-group. When do you think we can merge this? @rubenmoya.

rubenmoya:
That's in @javivelasco's hands!

package.json
```diff
@@ -143,7 +143,7 @@
     "prop-types": "^15.5.7",
     "react": "^15.5.0",
     "react-dom": "^15.5.0",
-    "react-transition-group": "^1.1.3"
```
rubenmoya: Why the mismatch between versions?

Merbs: Sorry for the delay; if you're asking why is this `2.2.0` and the other is `2.2.1`, I believe it's the way that yarn does it; maybe you think this is under `devDependencies` (because it looks that way on Github unless you expand it out), when really it's under  `peerDependencies`. If you look at classnames, there is a similar mismatch.

PR #1704, opened 2017-10-06
Fix GitHub TextCase.

potato4d:
The correct notation of GitHub is "GitHub".
But, since it was "Github" on the document, I fixed it.
`(s/Github/Github/g , exclude <GithubIcon />)`


potato4d:
@maxzitron Can you merge this?

PR #1717, opened 2017-10-17
Fix: Dropdown source item set as disabled throws warning

### LINKED ISSUE #1714: Dropdown source item set as disabled throws warning
Dropdown component throws warning when one of the source items has property disables as true.
When item is set as disabled, it works properly, but because of the PropTypes it throws a warning in console.

https://github.com/react-toolbox/react-toolbox/blob/dev/components/dropdown/Dropdown.js#L183

```
 Expected `onClick` listener to be a function, instead got a value of `boolean` type.
    in li (created by Dropdown)
    in ul (created by Dropdown)
    in div (created by Dropdown)
    in Dropdown (created by ThemedDropdown)
    in ThemedDropdown (created by Select)
```

Pajn:
fixes #1714

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -180,7 +180,7 @@ const factory = (Input) => {
         <li
           key={idx}
           className={className}
-          onMouseDown={!item.disabled && this.handleSelect.bind(this, item[valueKey])}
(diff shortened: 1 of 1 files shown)


SHGoldfarb:
It's been some time, what is preventing this from being merged?

PR #1720, opened 2017-10-19
Added support for keyboard navigation of dropdown menus

### LINKED ISSUE #1429: Dropdown accessibility
I have a dropdown with a custom template and it seems like tabbing into it does not work. When there is no template, it seems to work.

Additionally, the dropdown component doesn't let you use arrow keys to select different options.

sahilamoli:
Fixes #1429

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -1,12 +1,22 @@
 /* eslint-disable */
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import ReactDOM from 'react-dom';
+import { findDOMNode } from 'react-dom';
 import classnames from 'classnames';
 import { themr } from 'react-css-themr';
 import { DROPDOWN } from '../identifiers';
 import InjectInput from '../input/Input';
 import events from '../utils/events';
+import contains from 'dom-helpers/query/contains';
+import activeElement from 'dom-helpers/activeElement';
+import ownerDocument from 'dom-helpers/ownerDocument';
+
+const KEYS = {
+  UP_ARROW: 38,
+  DOWN_ARROW: 40,
+  TAB: 9,
+  ENTER: 13
+};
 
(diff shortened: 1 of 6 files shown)


pbflanagan:
@javivelasco

sahilamoli:
Let me know if any modifications are needed @javivelasco. Planning to submit more PRs after this gets merged :)

eL-HaXo:
When can we expect this PR to be merged and released?

components/dropdown/Dropdown.js
```diff
@@ -98,6 +112,38 @@ const factory = (Input) => {
...
+      let nextIndex = focusedItemIndex !== lastItemIndex ? focusedItemIndex + 1 : 0;
+
+      // If the next item is disabled then keep going until we either find a non-disabled item or we get back to the
+      // original focused item
+      while (source[nextIndex].disabled && nextIndex !== focusedItemIndex) {
+        nextIndex = nextIndex !== lastItemIndex ? nextIndex + 1 : 0;
+      }
+
+      return nextIndex;
+    };
```
fantactuka: Can be a little shorter:
```js
getNextSelectableItemIndex = (focusedItemIndex) => {
    const { source } = this.props;
    const lastItemIndex = source.length - 1;
    let nextIndex = focusedItemIndex;

    // If the next item is disabled then keep going until we either find a non-disabled item or we get back to the
    // original focused item
    do {
      nextIndex = nextIndex !== lastItemIndex ? nextIndex + 1 : 0;
    } while (source[nextIndex].disabled && nextIndex !== focusedItemIndex);

    return source[nextIndex];
}
```

guacamoli: @fantactuka, thanks. Not sure if this is getting merged anyway, so I'll wait on making any changes.

guacamoli:
@rubenmoya, can you take a look at this PR?

PR #1723, opened 2017-10-25
maxLength counter misaligned on Safari

### LINKED ISSUE #1440: Input Component: Misalignment of maxLength counter when error is rendered onFocus
FYI, this issue only occurs on Safari and Mobile Safari.

I am attempting to only show the error once my input has been focused once. In order to accomplish this, I use the state to determine when the input has been focused once and then I render the error. However, this causes the counter to be misaligned when the error is rendered after the input has been active. I have confirmed that this issue does not occur if the error is rendered prior to the input being active.

Repro'd using the react-toolbox playground:
```
class InputTest extends React.Component {
  state = { touched: false };

  handleChange = (name, value) => {
    this.setState({...this.state, [name]: value});
  };

  focused() {
    this.setState({
      touched: true
    });
  }
 
  getError() {
    if (this.state.touched) {
      return (<span>Error!!</span>);
    }
    return null;
  }

  render () {
    return (
      <section>
        <Input onFocus={this.focused.bind(this)} type='text' label='error' maxLength={20} error={this.getError()} />
      </section>
    );
  }
}

return <InputTest />;

```

On Chrome:
screen shot 2017-04-30 at 3 07 59 pm

On Safari:
screen shot 2017-04-30 at 3 07 36 pm

guacamoli:
-Fixes #1440
-Added a force repaint of the counter when an error is passed after the initial rendering
-Added test case in spec

### DIFF
--- a/components/input/Input.js
+++ b/components/input/Input.js
@@ -72,6 +72,12 @@ const factory = (FontIcon) => {
       type: 'text',
     };
 
+    // This state is only used to force repaint the counter
+    // @see https://github.com/react-toolbox/react-toolbox/issues/1440
+    state = {
+      forceSafariErrorRepaint: false,
+    };
+
     componentDidMount() {
       if (this.props.multiline) {
         window.addEventListener('resize', this.handleAutoresize);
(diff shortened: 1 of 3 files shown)


guacamoli:
@javivelasco

rubenmoya:
I think this is too much code for that kind of problem, I've been having a look at it and seems that applying some css to the counter fixes it.

```
position: absolute;
bottom: 2rem;
```

I just tried with your issues example, so it would be nice if you could give it a try in your project and see if that fixes it.

Thank you for the PR!

guacamoli:
@rubenmoya, yep. Looks like your fix works. I'll close this PR. Thanks!

PR #1725, opened 2017-10-26
Fix dropdown entry selection issue

rndevfx:
Currently a user have to click twice on an entry within the dropdown
list to close it. This happens because the li click event will call
'handleSelect' and the default behavior calls 'handleDocumentClick'.
This is wrong when the user clicks on an entry (and not outside of
the dropdown) so now the 'handleSelect' prevents the default behaviour.

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -99,6 +99,7 @@ const factory = (Input) => {
     };
 
     handleSelect = (item, event) => {
+      event.preventDefault()
       if (this.props.onBlur) this.props.onBlur(event);
       if (!this.props.disabled && this.props.onChange) {
         if (this.props.name) event.target.name = this.props.name;


rubenmoya:
Can you tell me the steps to reproduce that problem? I'm trying the dropdown [in the playground](http://react-toolbox.io/#/components/dropdown) and I'm not able to reproduce it :(

rndevfx:
Can you try it with the dev branch? On the website it works/worked fine but when you build the docs on your machine with latest dev (currently broken on my machine) you maybe can reproduce it :)

vitalif:
+1 it seems there's no such problem

louisremi:
I've had the same problem for quite some time now. I'm using react-toolbox with Preact and I assumed it was an incompatibility with that lib (see #1617 as well). I now assume the problem lies somewhere else, but putting `e.preventDefault` in my `onChange` handler did the trick.

PR #1732, opened 2017-10-31
Delete CNAME while domain is inactive

rubenmoya:


javivelasco:
Missed it, thank you @rubenmoya

PR #1737, opened 2017-11-02
Use window.requestAnimationFrame in Tabs.js

rndevfx:
Call requestAnimationFrame on window object to fix headless testing
with jsdom without hacking around by defining global vars


ss74:
Yup

PR #1755, opened 2017-11-13
Tooltip for disabled button

pablolmiranda:
Disabled buttons don't trigger mouseEnter and mouseLeave events, what creates an issue if you want to create a tooltip to inform why the button is disabled.
In this case we need to create a wrapper over the disabled button that will actually listen for the mouseEnter and mouseLeave events.

You can see the problem using the following playground code:
```
const TooltipButton = Tooltip(Button);
const TooltipInput = Tooltip(Input);
const TooltipLink = Tooltip(Link);

const TooltipTest = () => (
  <div>
    <TooltipButton label='Click on me' icon='bookmark' raised primary tooltip='Awesome, we can click' tooltipDelay={1000} />
    <TooltipButton label='Explain why you can click' icon='bookmark' raised primary tooltip='Where is the tooltip?' tooltipDelay={1000} disabled/>
  </div>
);

return <TooltipTest />;
```


components/button/Button.js
```diff
@@ -120,11 +120,22 @@ const factory = (ripple, FontIcon) => {
         'data-react-toolbox': 'button',
       };
 
-      return React.createElement(element, props,
+      const renderElementWrapper = element === 'button' && others.onMouseEnter && this.props.disabled;
```
rubenmoya: We shouldn't filter by element type `button` since the same problem happens with links too.

components/button/Button.js
```diff
@@ -120,11 +120,22 @@ const factory = (ripple, FontIcon) => {
...
+      const componentToRender = React.createElement(element, props,
         icon ? <FontIcon className={theme.icon} value={icon} /> : null,
         label,
         children,
       );
+
+      if (renderElementWrapper) {
```
rubenmoya: I'd do something like this, extracting the mouse events to a variable to reuse it and that allows you to return a ternary, but that's just code style.

```js
const mouseEvents = {
  onMouseUp: this.handleMouseUp,
  onMouseLeave: this.handleMouseLeave,
};

const props = {
  ...others,
  ...mouseEvents,
  href,
  ref: (node) => { this.buttonNode = node; },
  className: classes,
  disabled: this.props.disabled,
  type: !href ? type : null,
  'data-react-toolbox': 'button',
};

const buttonElement = React.createElement(element, props,
  icon ? <FontIcon className={theme.icon} value={icon} /> : null,
  label,
  children,
);

return others.onMouseEnter && this.props.disabled
  ? <span {...mouseEvents}>{buttonElement}</span>
  : buttonElement;
```

components/button/Button.js
pablolmiranda: I don't see the same happening for links. Could you give a code snippet showing that?

components/button/Button.js
pablolmiranda: Updated the PR with your code suggestions. Thank you for take a look on this.

rubenmoya:
About the problem in links, I just added a `href="#"` prop to the `<TooltipButton/>` in the website examples, it uses the same styles but renders an `a` element instead of `button`, so the `pointer-events: none` are applied and the tooltip doesn't show

PR #1761, opened 2017-11-21
Don't error when Tabs this.navigationNode changes soon

michaelfig:
Sometimes, this.navigationNode is set to null before the requestAnimationFrame callback runs.  This commit protects against that case.


michaelfig:
Clearly no interest in this PR.  This package is probably dead.

PR #1791, opened 2018-01-21
New prop proposal on Autocomplete component.

jesion:
I would like to propose a new prop on Autocomplete component - 'submitKeys'. It would mean an array of numbers representing keyboard keys that will trigger active item creation or selection.

This would allow programmers to configure which keyboard keys would trigger active item creation or selection. Defaults to 13, so this change is backwards compatible.

### DIFF
--- a/components/autocomplete/Autocomplete.js
+++ b/components/autocomplete/Autocomplete.js
@@ -54,6 +54,7 @@ const factory = (Chip, Input) => {
         values: PropTypes.string,
       }),
       value: PropTypes.any,
+      submitKeys: PropTypes.array
     };
 
     static defaultProps = {
@@ -67,6 +68,7 @@ const factory = (Chip, Input) => {
       showSuggestionsWhenValueIsSet: false,
       source: {},
       suggestionMatch: 'start',
+      submitKeys: [13]
     };
 
     state = {
@@ -145,9 +147,9 @@ const factory = (Chip, Input) => {
        && this.state.showAllSuggestions
      );
 
-      if (event.which === 13) {
-        this.selectOrCreateActiveItem(event);
-      }
+     if (this.props.submitKeys.includes(event.which) === true) {
+      this.selectOrCreateActiveItem(event);
+     }
     };
 
     handleQueryKeyUp = (event) => {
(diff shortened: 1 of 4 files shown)


components/autocomplete/Autocomplete.js
```diff
@@ -54,6 +54,7 @@ const factory = (Chip, Input) => {
         values: PropTypes.string,
       }),
       value: PropTypes.any,
+      submitKeys: PropTypes.array
```
3af: Maybe `PropTypes.arrayOf(PropTypes.number)`?

components/autocomplete/Autocomplete.js
```diff
@@ -145,9 +147,9 @@ const factory = (Chip, Input) => {
        && this.state.showAllSuggestions
      );
 
-      if (event.which === 13) {
-        this.selectOrCreateActiveItem(event);
-      }
+     if (this.props.submitKeys.includes(event.which) === true) {
```
3af: Is it necessary to compare to true? `includes()` returns boolean already.

components/autocomplete/readme.md
```diff
@@ -62,6 +62,7 @@ If you want to provide a theme via context, the component key is `RTAutocomplete
 | `showSuggestionsWhenValueIsSet` | `Bool`                        | `false` | If true, the list of suggestions will not be filtered when a value is selected, until the query is modified. |
 | `suggestionMatch`               | `String`                      | `start` | Determines how suggestions are supplied. It can be `start` (query matches the start of a suggestion), `anywhere` (query matches anywhere inside the suggestion), `word` (query matches the start of a word in the suggestion) or `disabled` (disable filtering of provided source, all items are shown). |
 | `value`                         | `String`, `Array` or `Object` |         | Value or array of values currently selected component. |
+| `submitKeys`                    | `Array`                       | `[13]`  | Array of numbers representing keyboard keys that will trigger active item creation or selection. |
```
3af: Maybe it would be helpful to link to resource where possible values are listed? E.g. https://css-tricks.com/snippets/javascript/javascript-keycodes/

jesion: Hi @3af, how about this one: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode?. Anyways, I have implemented the review changes...

3af: Hi @jesion, looks good too. Thanks for the changes.
Unfortunatelly I can't merge so let's wait till Javi has time for it.

jesion: No probs @3af. Thanks for reviewing...

Issue #1730, opened 2017-10-29
Disabled slider can be modified with keyboard input


Adamfsk:
On the demo page, if you use tab to select the disabled slider, then use the up and down arrows on the keyboard, the value of the disabled slider can be changed.

PR #1796, opened 2018-01-29
Don't handle key events if slider is disabled

rubenmoya:
Fixes: #1730


components/slider/Slider.js
```diff
@@ -10,6 +10,13 @@ import events from '../utils/events';
 import InjectProgressBar from '../progress_bar/ProgressBar';
 import InjectInput from '../input/Input';
 
+const KEYS = {
```
hartzis: i know `keyCode` was already being used here, but since you were cleaning it up thought i'd just point out that [keyCode is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and should probably be cleaned up in the future 😸 and replaced with [e.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)

rubenmoya: Good catch! Thank you!

hartzis: looks like you updated to use the `code` property but forgot to update the values in your `KEYS` map.
Should be
```
const KEYS = {
  ENTER: 'Enter',
  ESC: 'Escape',
  ARROW_UP: 'ArrowUp',
  ARROW_DOWN: 'ArrowDown',
};
```
Just for reference, good site for determining values for `code` - https://w3c.github.io/uievents/tools/key-event-viewer.html

rubenmoya: Ooops, what a fail 🤦🏻‍♂️

PR #1834, opened 2018-03-20
Set Slider tabIndex to -1 when disabled

dylanseago:
Set `tabIndex` to `-1` when the Slider is disabled to prevent it from gaining focus when navigating with a keyboard.


rubenmoya:
Thank you!

PR #1842, opened 2018-04-04
Adding a tabindex to the input component

thatgibbyguy:
When laying out a form, it's important to be able to control the order of the user's tabs. So, I added that as an optional prop that can be passed.


rubenmoya:
Hi! Sorry for answering this late, the code looks good, but the tests are failing

PR #1866, opened 2018-07-03
Add missing innerRef prop to typescript definition

rokoroku:
In `react-toolbox`, every components are wrapped by `react-css-themr`. 

Therefore, without `innerRef` from `react-css-themr`, we cannot get component's reference directly.

This PR will add the prop to the base `ReactToolbox.Prop` type


rubenmoya (approved):

rubenmoya:
Thank you!

PR #1869, opened 2018-07-14
using ref callback instead of settimeout to update clockNode reference

ankurarora:
Using callback to update ref and then call handleCalculateShape method from the reference


rubenmoya:
I tested it locally and seems to be working as expected, thank you for the PR!

arammkrtchyan:
Hi guys,
When are you planning the new release, which will contain this fix ?

PR #1884, opened 2018-09-28
Fix #1870: Dropdown won't close after selecting an item

### LINKED ISSUE #1870: Dropdown won't close after selecting an item
After replacing the onClick handler with onMouseDown in the Dropdown component (#1521, which fixed an IE11 issue), it needs two clicks to hide the selection list, although I can see the item was selected. (#1865 seems related too)

What I noticed looking into the props and prototypes of the Dropdown component is that **handleSelect** the **close** prototype is called, but also **handleFocus** is called again, which basically opens the selection list again.

jacargentina:
This PR fixes the problem when using the Dropdown **with mouse** at least. Maybe some other work should make it **work with the keyboard**, for example pressing the spacebar when the dropdown is focused, then accepting ENTER and ESC (to select, and close respectively)

@javivelasco @rubenmoya Please look when having some time. I can take your ideas to continue working on this.

### DIFF
--- a/components/dropdown/Dropdown.js
+++ b/components/dropdown/Dropdown.js
@@ -133,18 +133,6 @@ const factory = (Input) => {
       this.setState({ active: true, up });
     };
 
-    handleFocus = (event) => {
-      event.stopPropagation();
-      if (!this.props.disabled) this.open(event);
-      if (this.props.onFocus) this.props.onFocus(event);
-    };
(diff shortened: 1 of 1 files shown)


cav2008:
Hey I am experiencing the same problem with the dropdown. This fix looks like it works, will it get merged anytime soon?

PR #1890, opened 2018-10-25
fix css-transition animation for date & time picker

that1matt:
- fix css slide left & right transition animation for date picker
- fix css zoom in & out transition animation for time picker
- removed transition group extra div element
- remove extra div element on clock hours & clock minutes

### DIFF
--- a/components/time_picker/ClockHours.js
+++ b/components/time_picker/ClockHours.js
@@ -76,7 +76,7 @@ class Hours extends Component {
     const is24hr = format === '24hr';
 
     return (
-      <div>
+      <React.Fragment>
         <Face
           onTouchStart={this.handleTouchStart}
           onMouseDown={this.handleMouseDown}
@@ -98,7 +98,7 @@ class Hours extends Component {
           origin={center}
           step={step}
         />
-      </div>
+      </React.Fragment>
     );
   }
 }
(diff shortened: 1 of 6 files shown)


components/date_picker/theme.css
```diff
@@ -207,40 +207,44 @@
...
 
 .slideLeftEnter {
   opacity: 0;
-  transform: translate3d(-100%, 0, 0);
+  transform: translateX(-100%);
 
   &.slideLeftEnterActive {
     opacity: 1;
-    transform: translate3d(0, 0, 0);
+    transform: translateX(0);
```
rubenmoya: Why aren't we using `translate3d` anymore? It's supposed to be better since it relies on the GPU afaik

components/time_picker/ClockHours.js
```diff
@@ -76,7 +76,7 @@ class Hours extends Component {
     const is24hr = format === '24hr';
 
     return (
-      <div>
+      <React.Fragment>
```
rubenmoya, line 79: 😍

components/time_picker/ClockMinutes.js
```diff
@@ -41,7 +41,7 @@ class Minutes extends Component {
 
   render() {
     return (
-      <div>
+      <React.Fragment>
```
rubenmoya, line 44: 😍

components/date_picker/theme.css
that1matt: This was my bad, I reverted the change to use 3d.

rubenmoya (approved):

Issue #1918, opened 2019-06-11
ramda missing in dependencies


alexhisen:
"ramda": "^0.23.0" used to be in package.json's dependencies, but is now gone, so webpack compilation fails without it.

Issue #1913, opened 2019-02-27
Theming feature appears to be broken


rokoroku:
According to README.md file, theming feature can be achieved by `importFrom` feature from `postcss-custom-properties` module.

However, custom variables defined in `importFrom` does not take precedence above bundled css anymore (maybe since `postcss-custom-properties@8`) 
Take a look at: https://github.com/postcss/postcss-custom-properties/issues/148

So I'd suggest using `postcss-env-function` with `postcss-custom-variable` 
e.g.)
```css
:root {
  --primary-color: env(primary-color, var(--palette-indigo-500))
}
```

References:

https://developer.mozilla.org/en-US/docs/Web/CSS/env
https://github.com/postcss/postcss-custom-properties
https://github.com/jonathantneal/postcss-env-function

januszhou:
I'm seeing this issue as well, I had to install `"postcss-custom-properties": "^7.0.0"` and set `custom-properties: false`, I hope this helps.
```js
require('postcss-custom-properties')({
  preserve: false,
  variables: customProperties,
}),
require('postcss-preset-env')({
  stage: 0, // required to get all features that were from cssnext
     features: {
       'custom-properties': false,
       'color-mod-function': true, // required to use color-mod()
     },
 }),
```

PR #1920, opened 2019-06-25
dual-substitution process is not possible + updated documentation for Time Picker

neophytosd:
- Disabling the variable substitution and color-mod: We just can't get color-mod to preserve the original expression that still has variables and if you have to do color-mod compilation in your project, you might as well do only a single custom-properties substitution in your project like it always required (#1913).

- Updated the TimePicker documentation that it requires React 16.

- The gulpfile.js build file had incorrect paths for inlining @imports.

- Adds missing ramda dependency (#1918).

- We propose this be published as 3.0.0 due to the breaking changes in the build setup requirements.


alexhisen:
Please merge this PR to unbreak custom property overrides

alexhisen:
@rubenmoya, @that1matt, @javivelasco, @olegstepura Guys, it would be really great to merge these key PRs in!

PR #1885, opened 2018-10-08
Fixed issue of scrollbar being hidden on Autocomplete and Dropdown components

paddotk:
Dropdown and Autocomplete components had a `width: 0` and `height: 0` css rules. These don't serve any purpose and make users not know it's scrollable/there is more content, hence is bad user experience.

This fix takes out the two css lines to show scrollbar whenever necessary.

PR #1923, opened 2019-07-09
remove the unneeded css rules from Dropdown/Autocomplete ...

neophytosd:
... that break scrollability in those components


alexhisen:
This replaces #1885 which seems to have a lot of commits mixed into it.

